PR #8, opened 2024-08-21
chore: minor setup, vscode tweaks

haitmo:
Removed the msblack extension reference as it doesn't seem to exist (already installed through "ms-python.black-formatter")
changed debuger type as current one is being deprecated.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -6,7 +6,7 @@
     "configurations": [
         {
             "name": "Python: Run Tests",
-            "type": "python",
+            "type": "debugpy",
             "request": "launch",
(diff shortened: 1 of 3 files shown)


.vscode/launch.json
```diff
@@ -18,7 +18,7 @@
         },
         {
             "name": "Python: Run Court Monitor Tests",
```
juleskuehn, line 20: A bunch of these can be removed but that doesn't need to happen in this PR.

juleskuehn (dismissed):

juleskuehn (approved):

PR #9, opened 2024-08-21
ops: cloud shell prep

jannable:
The objectives of this PR as follows:

- [x] Adjust the instructions for them to run in Azure Cloud Shell
- [x] Simplify the creation of the .env file and add a confirmation step
- [x] Manage the terraform state more reliably regardless of user
- [x] Allow developers to continue to emulate within container
- [x] Fix issue with celery-worker pods, preventing k8s from succeeding
- [x] Set default OpenAI quotas and allow override
- [x] Reduce some prompts


setup/README.md
```diff
@@ -1,87 +1,98 @@
...
 
-The setup container contains all the necessary tools and dependencies to deploy the Otto infrastructure. You can choose to open the container in Visual Studio Code or directly in the Docker terminal.
+You have two options:
 
-**Option 1**: Open in Visual Studio Code
+**Option A: Clone the repository (if you haven't done so before)**
 
-Open Visual Studio Code and follow the steps below:
+```bash
+git clone https://github.com/justicecanada/Otto.git
```
juleskuehn, line 26: We renamed the repo "otto" lowercase.

setup/README.md
```diff
@@ -1,87 +1,92 @@
...
-The image for Otto needs to exist in the container registry. If the Terraform script just created it, the image will not exist on the registry yet. The development team will need to push the latest image to the registry first before continuing.
+## Setup Process
 
-### 1. Deploy the AKS cluster
+### 1. Clone the repository and navigate to the setup folder:
 
-Run the following command to deploy the AKS cluster:
+```bash
+git clone https://github.com/justicecanada/Otto.git
+cd Otto/setup
```
juleskuehn, line 72: Capitalization of "Otto" on lines 71 and 72

setup/README.md
```diff
@@ -1,87 +1,92 @@
...
-**Note:** The `CLIENT_SECRET` is not stored in the `.tfvars` file for security reasons. You can provide it interactively when prompted.
 
-### 2. Configure the OpenAI Capacity
+Important: Ensure the Otto image exists in the container registry before deploying the AKS cluster. (The image is built and pushed as part of the Continuous Integration pipeline.)
 
-After the Terraform script has completed deploying the OpenAI resource and its default model deployments, the capacity for each model deployment needs to be adjusted manually. This is necessary to ensure that the deployments have sufficient capacity to handle the expected workload.
+# Appendix: Development Team Guide
 
-**Note:** Instances of OpenAI within the same subscription share the same capacity pool. Ensure that the total capacity across all deployments does not exceed the subscription's capacity limit.
+For developers who want to emulate and test the scripts within a Docker container:
```
juleskuehn, line 58: What do you mean by "emulate"? Modify?
And they aren't really "testing" the scripts, they will be run just the same as if running in Cloud Shell, right? I think this is important because the statement makes it sound like it is totally safe to run them outside Cloud Shell and nothing can go wrong.

setup/README.md
```diff
@@ -1,87 +1,92 @@
-# Otto Setup Guide
+# Otto Setup Guide for Cloud Admins
```
juleskuehn, line 1: All references to the .env have been removed which I think is an issue

juleskuehn (approved):
I will approve so you can merge but I have some comments that should probably be addressed.

setup/README.md
jannable: They can be run from outside the Cloud Shell though.

setup/README.md
jannable: The script prompts for the values and shows the user.

juleskuehn (approved):

PR #12, opened 2024-08-22
feat: cost tracking (per user, per feature)

juleskuehn:
Langchain token counter was giving me trouble so I have decided to remove Langchain from AI assistant.

Costs have this granularity:
- Feature (one of the otto features: Chat, Summarize, Translate, Laws, Template Wizard, etc)
- CostType (e.g. "GPT-3.5-in" or "embedding-query" - these are defined in cost_types.yaml fixture)
- User (optional, but presently tracked in all cases. Optional so that scheduled data refreshes do not need to be attributed to a particular user)

Tasks:
- [x] Costing models
- [x] LlamaIndex wrapper class to help with costing, reduce boilerplate
- [x] Refactor AI assistant to remove Langchain
- [x] Record costs for AI assistant LLM & embedding calls, including titling chats (recorded as "chat" feature always)
- [x] Record costs for file translation
- [x] Record costs for librarian (embedding)
- [x] Refactor htmx_stream, chat message templates (DRY)
- [x] Display costs in AI assistant messages
- [x] Display costs in librarian
- [x] Display costs in manage users dashboard
- [x] Record Azure Document AI costs (PDF loading during summarize file or librarian - if Fitz fails)
- [x] Edit tests and/or new tests

Maybe in separate PRs?
- [ ] Record costs for Laws search
- [ ] Record costs for OCR tool
- [ ] Rethink / fix stop button behaviour (doesn't return token usage; doesn't actually conserve tokens either!)
- [ ] Cost stopped Celery tasks

Out of scope for now:
* Case prep assistant (currently under rework)
* Template wizard (soon to be reworked; would need a major refactor to remove Langchain)

**Questions**
1. Should "feature" really be linked to the otto.models.Feature model? It may be more flexible to simply have an enum of strings (which include all the features, but can also support more misc. things like "chat-title")

File translation response (Azure Translation costs):
image

Chat response (LLM input and output token costs):
image

QA response (LLM tokens + query embedding costs):
image

Summarizing a long file (LLM token costs):
image

In Librarian modal:
image

Total cost per user (in "manage users"):
image


django/locale/translation/translations.json
```diff
@@ -3218,5 +3218,9 @@
     "Maximum sources": {
         "fr": "",
         "fr_auto": "Sources maximales"
+    },
+    "Otto": {
+        "fr": "Otto",
+        "fr_auto": "Otton"
```
jannable, line 3224: French auto has "n"

juleskuehn: Gets overriden by "fr" value (manually added to address this problem).

django/chat/llm.py
```diff
@@ -0,0 +1,252 @@
...
+    @property
+    def output_token_count(self):
+        return self._token_counter.completion_llm_token_count
+
+    @property
+    def embed_token_count(self):
+        return self._token_counter.total_embedding_token_count
+
+    def create_costs(
+        self, user: User, feature: str, embed_task: str = "query"
```
juleskuehn, line 110: to refactor: get rid of "embed_task". instead, make a feature "librarian".
| | Current | Proposed |
|--------|--------|--------|
| **feature** | qa | librarian |
| **cost_type** | embed-doc | embed |

| | Current | Proposed |
|--------|--------|--------|
| **feature** | qa | qa |
| **cost_type** | embed-query | embed |

django/chat/llm.py
```diff
@@ -0,0 +1,252 @@
...
+from llama_index.core.retrievers import QueryFusionRetriever
+from llama_index.core.vector_stores.types import MetadataFilter, MetadataFilters
+from llama_index.embeddings.azure_openai import AzureOpenAIEmbedding
+from llama_index.llms.azure_openai import AzureOpenAI
+from llama_index.vector_stores.postgres import PGVectorStore
+
+from otto.models import Cost, User
+
+
+class OttoLLM:
```
juleskuehn, line 16: "LLM" is a slight misnomer since this also includes embedding / vector store functions, but I wasn't sure what else to call it. And it is far more convenient to have them bundled together here.

django/chat/responses.py
```diff
@@ -55,25 +51,32 @@ def otto_response(request, message_id=None):
 
 def chat_response(chat, response_message, eval=False):
 
-    user_message = response_message.parent
+    def is_text_to_summarize(message):
+        return message.mode == "summarize" and not message.is_bot
 
     system_prompt = chat.options.chat_system_prompt
-    chat_history = [SystemMessage(content=system_prompt)]
+    chat_history = [ChatMessage(role=MessageRole.SYSTEM, content=system_prompt)]
```
juleskuehn, line 58: Refactored to use LlamaIndex throughout

django/chat/responses.py
```diff
@@ -55,25 +51,32 @@ def otto_response(request, message_id=None):
...
-        (
-            AIMessage(content=message.text)
-            if message.is_bot
-            else HumanMessage(content=message.text)
+        ChatMessage(
+            role=MessageRole.ASSISTANT if message.is_bot else MessageRole.USER,
+            content=(
+                message.text
+                if not is_text_to_summarize(message)
+                else "<text to summarize...>"
```
juleskuehn, line 65: This is to prevent long texts (which user input for summarization purposes) to be passed into Chat mode later, if the mode is switched from Summarize to Chat.

django/chat/responses.py
```diff
@@ -141,54 +134,76 @@ def summarize_response(chat, response_message):
     target_language = chat.options.summarize_language
     model = chat.options.summarize_model
 
+    llm = OttoLLM(model)
+
+    # TODO: Extracting text from file may incur Azure Document AI costs.
+    # Need to refactor extract_text to create Cost object with correct user and mode.
```
juleskuehn, line 140: It is annoying to do all the refactoring to pass around user and feature into any function that incurs costs. Often the function in which you could actually calculate costs is several layers deep.

I would like to refactor this so that there is some middleware etc. that keeps track of the request.user and feature for each request in a common cache that any function can access. Because Django processes each request in its own thread, this is possible. It is possible to mess up with async functions, and probably Celery also complicates matters.

django/chat/responses.py
```diff
@@ -141,54 +134,76 @@ def summarize_response(chat, response_message):
...
     async def multi_summary_generator():
+        full_text = ""
         for i, file in enumerate(files):
-            yield f"**{file.filename}**\n\n"
+            full_text += f"**{file.filename}**\n\n"
+            yield full_text
             if not file.text:
                 await sync_to_async(file.extract_text)(fast=True)
-            summary = await summarize_long_text_async(
+            response_stream = await summarize_long_text_async(
```
juleskuehn, line 148: Streaming summaries now, even for multiple file uploads.

This can be parallelized but I didn't want to get into it right now. The idea would be combining multiple streaming responses (for different documents or chunks). I have discussed this with Shaun to implement in Advanced Q&A since it is highly applicable there.

django/chat/responses.py
```diff
@@ -89,27 +92,17 @@ def chat_response(chat, response_message, eval=False):
...
 
-    llm_stream = llm.astream(chat_history)
     return StreamingHttpResponse(
         streaming_content=htmx_stream(
             chat,
             response_message.id,
-            response_stream=llm_stream,
-            chunk_property="content",
+            response_replacer=llm.chat_stream(chat_history),
+            llm=llm,
```
juleskuehn, line 105: The LLM object is passed in here because its costing methods need to be called *after* the response is complete. Otherwise there would be no reference to the object from which to access the token counts.

django/chat/responses.py
```diff
@@ -245,8 +259,8 @@ async def file_translation_generator(task_ids):
                 chat,
                 response_message.id,
                 response_replacer=file_translation_generator(task_ids),
+                dots=True,
                 format=False,  # Because the generator already returns HTML
-                save_message=False,  # Because the generator already saves messages
```
juleskuehn, line 249: Due to costing being added (etc), htmx_stream always saves the messages now.

django/chat/static/chat/scripts.js
```diff
@@ -5,9 +5,9 @@ const copyCodeButtonHTML = `<button type="button" onclick="copyCode(this)"
 class="btn btn-link m-0 p-0 text-muted copy-message-button copy-button"
 title="Copy"><i class="bi bi-clipboard"></i><i class="bi bi-clipboard-fill"></i></button>`;
 
-function scrollToBottom(smooth = true) {
+function scrollToBottom(smooth = true, force = false) {
   resizePromptContainer();
-  if (preventAutoScrolling) {
+  if (preventAutoScrolling && !force) {
```
juleskuehn, line 10: Force the chat window to scroll to the bottom when a new user message is submitted.

django/chat/tasks.py
```diff
@@ -65,11 +67,17 @@ def translate_file(file_path, out_message_id, target_language):
...
         result = poller.result()
 
-        # Estimate the cost
-        # usage = poller.details.total_characters_charged
-        # estimated_cost = calculate_cost(usage, settings.AZURE_DOCUMENT_TRANSLATION)
-
         out_message = Message.objects.get(id=out_message_id)
+        usage = poller.details.total_characters_charged
+        cost = Cost.objects.new(
+            user=out_message.chat.user,
```
juleskuehn, line 73: Cost created within a Celery task.
Because we have passed in the message_id, we can retrieve the user from there.
We would have to do something like this for any Celery task in order to have the information we need for costing.

I feel like the Cost.objects.new() method is succinct enough to not need to abstract it further.

django/chat/tasks.py
```diff
@@ -65,11 +67,17 @@ def translate_file(file_path, out_message_id, target_language):
...
+            feature="translate",
+            cost_type="translate-doc",
+            count=usage,
+        )
+        print(f"Cost: {cost}")
+        out_message.cost = float(out_message.cost) + float(cost.usd_cost)
```
juleskuehn, line 79: I also need to save the total costs inside the message, in order to display the cost per message to the user.
This might actually sum multiple cost objects, either of the same type (like translating multiple files, here) or of multiple types (embedding, in tokens, out tokens)

django/chat/templates/chat/components/chat_message.html
```diff
@@ -1,8 +1,9 @@
 {% load static %}
 {% load i18n %}
 
-<div class="message-outer {% if message.is_bot %}bot{% elif not file_upload and not message.num_files %}user{% endif %}{% if message.awaiting_response %} awaiting-response{% endif %}"
-     id="message_{{ message.id }}">
+<div class="message-outer {% if message.is_bot %}bot{% elif not file_upload and not message.num_files %}user{% endif %} {% if message.awaiting_response %}awaiting-response{% endif %} {% if swap_oob %}show-all{% endif %}"
+     id="message_{{ message.id }}"
+     {% if swap_oob %}hx-swap-oob="true"{% endif %}>
```
juleskuehn, line 6: As I mentioned earlier, this template is now rendered at the end of htmx_stream, which ends the streaming response (since the outer message, which previously contained the SSE element, has been replaced with one that does not have a SSE element).

django/chat/templates/chat/components/streaming_response.html
```diff
@@ -1,7 +1,6 @@
 <div hx-ext="sse"
      sse-connect="{% url 'chat:chat_response' message_id=message.id %}"
      sse-swap="message"
-     hx-swap="morph:innerHTML"
```
juleskuehn, line 4: Morphing stopped working at some point. I only really had it so that the dots would continue their animation smoothly. Very low priority to fix.

django/chat/utils.py
```diff
@@ -94,240 +97,168 @@ def url_to_text(url):
...
+        full_message = _("An error occurred:") + f"\n```\n{str(e)}\n```"
+        message.text = full_message
+        await sync_to_async(message.save)()
+        message.text = llm_response_to_html(full_message)
+        context = {"message": message, "swap_oob": True}
+
+    # Render the message template, wrapped in SSE format
+    yield sse_string(
+        await sync_to_async(render_to_string)(
+            "chat/components/chat_message.html", context
```
juleskuehn, line 251: This kills several birds with one stone. Stop the SSE response. Present the final message including costs. Update the chat title. Update the chat security label.

django/chat/utils.py
```diff
@@ -357,253 +288,90 @@ def title_chat(chat_id, force_title=True):
...
-            return await map_reduce_chain.arun(split_docs)
+    # Tree summarizer prompt requires certain variables
+    # Note that we aren't passing in a query here, so the query will be empty
+    length_prompt_template = length_prompt_template.replace(
+        "{docs}", "{context_str}{query_str}"
+    )
+    template = PromptTemplate(length_prompt_template, prompt_type=PromptType.SUMMARY)
 
-        response = asyncio.run(run_map_reduce())
+    response = llm.tree_summarize(
```
juleskuehn, line 360: Simpler than with previous Langchain implementation. This will only make multiple LLM calls if the input tokens exceed the model's maximum (in case of GPT-4o, 128k). This very rarely happens, so you get the response sooner. It is also cheaper without the additional "reduce" calls to summarize the chunk summaries. I haven't noticed a drop in quality vs. the map reduce summaries but also haven't tested much.

django/chat/views.py
```diff
@@ -369,10 +375,10 @@ def chat_message(request, chat_id):
     user_message_text = request.POST.get("user-message", "").strip()
     mode = chat.options.mode
 
-    logger.info(
+    logger.debug(
         "User message received.",
         chat_id=chat_id,
-        user_message=user_message_text,
+        user_message=f"{user_message_text[:100]}{'...' if len(user_message_text) > 100 else ''}",
```
juleskuehn, line 381: Logger was printing the entire message which could get really crazy if you were entering a whole document for summarization etc.

django/librarian/models.py
```diff
@@ -137,10 +136,10 @@ def delete(self, *args, **kwargs):
         self.reset(recreate=False)
         super().delete(*args, **kwargs)
 
-    def process_all(self, force=True):
+    def process_all(self, force=True, user=None):
```
juleskuehn, line 139: This is the kind of changes to "functions called from inside a request that incur costs" that I'd like to avoid.

django/librarian/models.py
```diff
@@ -377,15 +382,16 @@ def delete(self, *args, **kwargs):
...
         from .tasks import process_document
 
         # Logic for updating the document embeddings, metadata, etc.
         if not (self.file or self.url):
             self.status = "ERROR"
             self.save()
             return
-        process_document.delay(self.id, get_language())
+        # User is optionally passed through for cost tracking purposes
+        process_document.delay(self.id, user.upn if user else None, get_language())
```
juleskuehn, line 394: There won't always be a user to attribute the cost to. e.g. when loading documents from a setup script, or scheduled refresh.

django/librarian/tasks.py
```diff
@@ -18,15 +19,15 @@
...
+from otto.models import User
 
 logger = get_logger(__name__)
 
 ten_minutes = 600
 
 
 @shared_task(soft_time_limit=ten_minutes)
-def process_document(document_id, language=None):
+def process_document(document_id, user_upn=None, language=None):
```
juleskuehn, line 30: Unlike the file translation Celery function, I had to pass a string for the user in here. 
You can't pass User objects directly since they aren't serializable.
And we can't infer the user from the Document since more than 1 user can have responsibility for a document.data_source.library.

django/librarian/tasks.py
```diff
@@ -41,17 +42,23 @@ def process_document(document_id, language=None):
...
+        document.status = "ERROR"
+        document.celery_task_id = None
+        document.save()
 
-        except SoftTimeLimitExceeded:
-            document.status = "ERROR"
-            document.celery_task_id = None
-            document.save()
+    user = User.objects.get(upn=user_upn) if user_upn else None
+    costs = llm.create_costs(user, feature="qa", embed_task="doc")
```
juleskuehn, line 56: This feature would change to "librarian", with no "embed_task" argument, if I refactor as planned.

django/librarian/tasks.py
```diff
@@ -41,17 +42,23 @@ def process_document(document_id, language=None):
...
+    document.cost = costs[0].usd_cost
+    document.save()
 
 
-def process_document_helper(document):
+def process_document_helper(document, llm):
```
juleskuehn, line 61: Again, "llm" is a slight misnomer, but this is the OttoLLM object that includes the vector store helpers and costing stuff.

django/otto/fixtures/apps.yaml
```diff
@@ -7,27 +7,31 @@
     user_group: AI assistant user
     features:
       - fields:
+          short_name: "qa"
```
juleskuehn, line 10: I was using these to more easily attribute Costs to particular Features.
However I think I would rather get rid of these, and not have a direct relationship between these Feature objects and Costs. This would be more flexible, e.g. for adding "librarian" feature which does not actually exist as a "Feature" in Otto.

django/otto/fixtures/cost_types.yaml
```diff
@@ -0,0 +1,160 @@
...
+    description: OpenAI Embedding 3 Large model for processing documents
+    description_fr: Modèle OpenAI Embedding 3 Large pour le traitement des documents
+    unit_name: tokens
+    unit_name_fr: jetons
+    unit_cost: 0.00013
+    unit_quantity: 1000
+- model: otto.CostType
+  pk: 8
+  fields:
+    short_name: embed-query
```
juleskuehn, line 90: embed-query and embed-doc can be combined if "librarian" is a feature.

django/otto/models.py
```diff
@@ -237,3 +245,87 @@ def maximum_of(cls, acronyms):
...
+            cost_type=cost_type,
+            count=count,
+            usd_cost=(count * cost_type.unit_cost) / cost_type.unit_quantity,
+        )
+
+    def get_user_cost(self, user):
+        # Total cost for a user
+        return sum(cost.usd_cost for cost in self.filter(user=user))
+
+    def get_user_cost_by_type(self, user, cost_type):
```
juleskuehn, line 285: Examples of the kinds of breakdowns we might want to show users or Otto admins.

django/otto/settings.py
```diff
@@ -91,18 +91,17 @@
 # OpenAI
 AZURE_OPENAI_ENDPOINT = os.environ.get("AZURE_OPENAI_ENDPOINT")
 AZURE_OPENAI_VERSION = os.environ.get("AZURE_OPENAI_VERSION")
-# https://openai.com/pricing
+# TODO: Replace with Cost model (this is used in Template Wizard)
 OPENAI_COST_PER_TOKEN = 0.0020 / 1000
 OPENAI_EMBEDDING_COST_PER_TOKEN = 0.0004 / 1000
 
 DEFAULT_CHAT_MODEL = "gpt-4o"
+USD_TO_CAD = 1.36
```
juleskuehn, line 99: Need to keep this updated (maybe once per year).

django/chat/responses.py
```diff
@@ -31,12 +33,17 @@
...
+    Stream a response to the user's message. Uses LlamaIndex to manage chat history.
     """
     response_message = Message.objects.get(id=message_id)
     chat = response_message.chat
     assert chat.user_id == request.user.id
     mode = chat.options.mode
+
+    # For costing and logging. Contextvars are accessible anytime during the request
+    # including in async functions (i.e. htmx_stream) and Celery tasks.
+    bind_contextvars(message_id=message_id, feature=mode)
```
juleskuehn, line 45: By adding to the structlog contextvars, we can use this information in the Cost objects whenever they are created (including from async functions or from Celery tasks)

django/chat/utils.py
```diff
@@ -212,12 +212,9 @@ def sse_string(message: str, format=True, dots=False, remove_stop=False) -> str:
...
         yield sse_string(full_message, format, dots=False, remove_stop=True)
 
-        # Response is done! Update costs and save message
+        await sync_to_async(llm.create_costs)()
+
         message = await sync_to_async(Message.objects.get)(id=message_id)
-        user = await sync_to_async(lambda: chat.user)()
-        costs = await sync_to_async(llm.create_costs)(user, message.mode)
-        total_cost = sum([cost.usd_cost for cost in costs])
-        message.cost = total_cost
```
juleskuehn: The message and document cost properties are now calculated / set from within the Cost.objects.new() function.

django/librarian/models.py
```diff
@@ -153,7 +157,7 @@ def reset(self, recreate=True):
         session.close()
         if recreate:
             # This will create the vector store table
-            llm.get_index(self.uuid_hex)
+            llm.get_retriever(self.uuid_hex).retrieve("?")
```
juleskuehn, line 160: Bug fix. llm.get_index() was not actually creating the database table. However, querying a blank index does actually create the table. This is important because otherwise, multiple Celery tasks may try to create the table simultaneously.

django/librarian/templates/librarian/components/document_status.html
```diff
@@ -109,8 +109,8 @@ <h6>{% trans "Document status" %}</h6>
           {{ document.num_chunks }}
         </dd>
       {% endif %}
-      {% if document.cost %}
-        <dt>{% trans "Processing cost" %}</dt>
+      {% if document.usd_cost %}
+        <dt>{% trans "Processing cost (cumulative)" %}</dt>
```
juleskuehn, line 113: Because of the way costs are calculated / set for the documents, only cumulative cost is really possible now.

django/librarian/utils/process_engine.py
```diff
@@ -514,7 +518,7 @@ def _pdf_to_html_using_azure(content):
     result = poller.result()
 
     num_pages = len(result.pages)
-    # TODO: create cost object based on number of pages
+    cost = Cost.objects.new(cost_type="doc-ai-prebuilt", count=num_pages)
```
juleskuehn, line 521: Form recognizer aka Document AI cost. This works regardless of how this is invoked, i.e. file summarization or librarian upload.

django/otto/models.py
```diff
@@ -267,17 +267,41 @@ def __str__(self):
 
 
 class CostManager(models.Manager):
-    def new(self, user: User, feature: str, cost_type: str, count: int) -> "Cost":
-        feature = Feature.objects.get(short_name=feature)
+
+    def new(self, cost_type: str, count: int) -> "Cost":
+        from structlog.contextvars import get_contextvars
```
juleskuehn, line 272: This is the guts of the refactor. This method now only takes two required arguments. The rest are populated from contextvars, if available.

django/otto/models.py
```diff
@@ -314,18 +338,48 @@ def get_user_cost_today(self, user):
...
+    ("qa", _("Q&A")),
+    ("chat", _("Chat")),
+    ("translate", _("Translate")),
+    ("summarize", _("Summarize")),
+    ("template_wizard", _("Template wizard")),
+    ("laws_query", _("Legislation search")),
+    ("laws_load", _("Legislation loading")),
+    ("case_prep", _("Case prep assistant")),
+    ("text_extractor", _("Text extractor")),
+]
```
juleskuehn, line 352: We can extend this later, but it will result in a (harmless, if only adding choices) migration.

django/otto/models.py
```diff
@@ -314,18 +338,48 @@ def get_user_cost_today(self, user):
...
+    usd_cost = models.DecimalField(max_digits=12, decimal_places=6)
+
+    # Optional, for aggregation and reporting
+    user = models.ForeignKey(User, on_delete=models.CASCADE, null=True, blank=True)
+    feature = models.CharField(
+        max_length=50, null=True, blank=True, choices=FEATURE_CHOICES
+    )
+
+    # Optional, for debugging purposes
+    request_id = models.CharField(max_length=50, null=True, blank=True)
```
juleskuehn, line 373: Matches the request_id in the logs

django/otto/utils/common.py
```diff
@@ -13,7 +13,18 @@ def file_size_to_string(filesize):
...
+    if approx_cost_cad < 0.01:
         return "< $0.01"
     return f"${approx_cost_cad:.2f}"
+
+
+def set_costs(object):
+    """
+    Sums cost.usd_cost from the object's cost_set and assigns total to object.usd_cost
+    """
+    object.usd_cost = sum([cost.usd_cost for cost in object.cost_set.all()])
```
juleskuehn, line 29: Costs will now always be cumulative for librarian documents.

haitmo (approved):
Looks good thanks Jules!

PR #15, opened 2024-08-26
ops: Celery fix

jannable:


django/initial_setup.sh
```diff
@@ -1,5 +1,8 @@
 #!/bin/sh
 
+echo "Installing Python dependencies..."
+pip install -r requirements.txt || { echo "Error: Pip install failed"; exit 1; }
```
haitmo, line 4: This is taken care of in the docker file already no?

setup/.env.example
```diff
@@ -1,31 +1,87 @@
...
-CRITICALITY=NonEssential # Criticality of the application (e.g. Essential, NonEssential)
-OWNER="Business Analytics Centre" # Owner of the application
-
-# Application
-DJANGO_ENV=DEV # Django environment (e.g. DEV, UAT, PROD)
-DJANGO_DEBUG=True # Django debug mode (e.g. True, False)
-
-# OpenAI Usage Capacity
+# Version number for this environment file
+ENV_VERSION=1.2
```
haitmo, line 2: Shouldn't this be 1.1?

setup/.env.example
```diff
@@ -1,31 +1,87 @@
...
+### APPLICATION URL ###
+
+# Option 1: Custom Domain (Production)
+# Uncomment and set this for environments with a custom domain
+# SITE_URL=https://otto.justice.gc.ca
+
+# Option 2: Azure-provided Domain
+# Uncomment and set this for environments without a custom domain
+# Format: otto-<environment>-<uniqueidentifier>
+DNS_LABEL=otto-sandbox-abc123
```
haitmo, line 29: What is the uniqueidentifier for?

setup/.env.example
```diff
@@ -1,31 +1,87 @@
...
+### APPLICATION VARIABLES ###
+
+# Django environment (e.g. DEV, UAT, PROD)
+DJANGO_ENV=DEV 
+
+# Django debug mode (e.g. True, False)
+DJANGO_DEBUG=True
+
+# Application admin (comma-separated list of UPNs)
+OTTO_ADMIN=jason.annable@justice.gc.ca,hamza.aitmohamed@justice.gc.ca,jules.kuehn@justice.gc.ca
```
haitmo, line 67: Can we leave off email addresses from the repo and provide an example of an email address?

haitmo (dismissed):

django/initial_setup.sh
jannable: Oops, meant to remove it earlier.

setup/.env.example
jannable: This is a public repo and it's conceivable someone else might try to stand this up in the same canadacentral region. the unique identifier is to cover the theoretical possibility of a conflict.

haitmo (approved):

PR #23, opened 2024-08-29
chore: update the docker container setup section of the README

mcusteau:
step 8 of the docker container setup section is outdated. It needs to be updated to use the set_admin_user command


haitmo (approved):
Thanks Michel!

PR #26, opened 2024-09-04
feat: skip to main content

mcusteau:
TICKET 2705 https://dev.azure.com/BusinessAnalyticsCentre/BAC%20Agile/_workitems/edit/2705

Adds a skip to main content button thats hidden unless focused on. The button is on the home page and on each app (except for template wizard as this one is getting redesigned)


mcusteau:
TICKET 2705 https://dev.azure.com/BusinessAnalyticsCentre/BAC%20Agile/_workitems/edit/2705

Adds a skip to main content button thats hidden unless focused on. The button is on the home page and for each app (except for template wizard as this one is getting redesigned)

django/otto/templates/base.html
mcusteau, line 1: you will notice I moved {% include 'components/search.html' %} down a few lines to be after the Otto logo and the skip navigation button. This doesn't change the way it appears on the Otto website. The only thing it changes is that when the user presses the Tab key, instead of being taken directly to the search bar, they are first taken to the logo of the website, the skip navigation button, and then the search bar. I checked and this is how major websites like Youtube and Reddit do it.

juleskuehn (changes requested):
Hey Michel, I can confirm this is working!

Can you please put the skip link as the first interactive HTML element, and ensure it never gets behind the search bar? (may need to adjust z-index). Right now it can sometimes hide behind depending on viewport size:
image

I appreciate that you looked into Reddit and YouTube! I like Reddit's implementation. Although YouTube puts it 3rd, I would prefer the skip link come first in the tab order, as that is more common and will make a good first impression for the a11y testers.

Can you add these classes so it looks more like a button and sticks to the left? (Just my preference.) It is OK for it to go overtop of the logo, like on Reddit. Also helps to avoid potential overlapping the search field.
```
btn btn-primary position-absolute
```
image

Also, you should only have to change base.html, not the other ones, so all that main_content_id block stuff can be avoided. This will make it easier to maintain.
e.g.
```html
    <div id="main-content-wrapper">
      {% block content_container %}
        <div class="container body-content">
          <main class="py-md-5 py-3" role="main">
            {% block body %}{% endblock %}
          </main>
        </div>
      {% endblock %}
    </div>
```

And the link can then just always go to that element, e.g.
```
      <nav class="navbar bg-dark border-bottom navbar-expand-sm position-relative"
           data-bs-theme="dark"
           id="top-nav">
        <a href="#main-content-wrapper"
            class="skip-to-main-content visually-hidden-focusable">{% trans "Skip to main content" %}</a>
```

Thanks!

juleskuehn (approved):

PR #35, opened 2024-09-09
fix: celery gevent pool, 1000 concurrency

juleskuehn:
Works locally with 1000 tasks launched simultaneously.
Needs to be tested in AKS.

Easiest/cheapest way to launch a lot of concurrent tasks is to upload a ton of small files in librarian.


haitmo:
A few things I wanted to try:
- Setting a cap for memory 
- Setting a cap for tasks before the process being replace
- Using autoscale (probably for cpu bound tasks)

command: ['celery', '-A', 'otto', 'worker', '-l', 'INFO', '--autoscale=8,1', '--max-memory-per-child=250000', '--max-tasks-per-child=1']

juleskuehn:
> --max-tasks-per-child=1

Wouldn't this do the opposite of what we want? Or does it only apply to children rather than the driver node?

PR #36, opened 2024-09-09
fix: correct behaviour for advanced q+a prompt inputs

smcdonald-jus:
Advanced Q+A options didn't register backspaces or retain the last character of user input. Both were fixed by changing keypress triggers to keyup triggers.


juleskuehn (approved):
Thanks for catching this! Nice quick fix.

PR #38, opened 2024-09-09
feat: per-source Q&A

smcdonald-jus:


django/chat/utils.py
```diff
@@ -393,3 +393,28 @@ async def summarize_long_text_async(
     return await sync_to_async(summarize_long_text)(
         text, llm, length, target_language, custom_prompt
     )
+
+
+async def combine_responses(responses, sources):
```
juleskuehn, line 398: I played with this a bit and prefer this formatting. (Also got rid of the comment)
```
async def combine_responses(responses, sources):
    streams = [
        {"stream": stream.response_gen, "status": "running"} for stream in responses
    ]
    final_streams = [
        f"\n###### *{source.metadata.get('title', source.metadata['source'])}*\n"
        for source in sources
    ]
    while any([stream["status"] == "running" for stream in streams]):
        for i, stream in enumerate(streams):
            try:
                if stream["status"] == "running":
                    final_streams[i] += next(stream["stream"])
            except StopIteration:
                stream["status"] = "stopped"
        yield ("\n\n---\n\n".join(final_streams))
        await asyncio.sleep(0)

```
If this looks OK to you, please substitute before merging.

django/chat/responses.py
```diff
@@ -425,19 +426,39 @@ async def add_files_to_library():
             content_type="text/event-stream",
         )
 
-    response = synthesizer.synthesize(query=input, nodes=source_nodes)
+    if chat.options.qa_answer_mode != "per-source":
+        response = synthesizer.synthesize(query=input, nodes=source_nodes)
+        response_generator = response.response_gen
+        sources = response.source_nodes
+        return StreamingHttpResponse(
```
juleskuehn, line 433: You could probably refactor slightly so "sources" is only defined in 1 place, and the same return statement is used. This would DRY it out even more.

Just pass response_generator=None in the per-source case, and response_replacer=None in the normal case.

juleskuehn (approved):
Great! That is nice and simple. Please:
* address the two comments if possible / you agree
* add a test which exercises this ChatOption

Then feel free to merge. Thanks!

django/chat/utils.py
smcdonald-jus: You're right - this is much nicer!

juleskuehn (approved):

PR #41, opened 2024-09-11
refactor(laws): autocomplete multiselect for laws advanced search

juleskuehn:


django/chat/llm.py
```diff
@@ -43,6 +47,7 @@ def __init__(
         )
         self._callback_manager = CallbackManager([self._token_counter])
         self.llm = self._get_llm()
+        self.mock_embedding = mock_embedding
```
juleskuehn, line 50: The Mock embedding is for costing or testing purposes. e.g. skipping embedding a bunch of a laws when I'm trying to test how they are loaded. We should add a Mock LLM here later as well, for estimating costs before running expensive processes.

django/laws/forms.py
```diff
@@ -0,0 +1,244 @@
...
+                {"label": str(x), "value": str(x.id)}
+                for x in data
+                if str(x.id) in values
+            ]
+            return items
+
+        return []
+
+
+class LawSearchForm(forms.Form):
```
juleskuehn, line 69: The main work of this PR is the new advanced search form.

django/laws/management/commands/load_laws_xml.py
```diff
@@ -680,180 +755,283 @@ def handle(self, *args, **options):
...
+                document_fr,
+                nodes_fr,
+                sha_256_hash_en=en_hash,
+                sha_256_hash_fr=fr_hash,
+                add_to_vector_store=True,
+                force_update=force_update,
+                llm=llm,
+            )
+            bind_contextvars(law_id=law.id)
+            llm.create_costs()
```
juleskuehn, line 1029: Costs for embedding laws are saved and associated with particular Law object.

django/laws/views.py
```diff
@@ -82,34 +86,42 @@ def _get_other_lang_node(node_id):
     return _get_source_node(other_lang_node_id)
 
 
-def _get_law_url(law):
+def _get_law_url(law, request_lang):
```
juleskuehn, line 89: Law objects are now Bilingual so some refactoring was required.

django/laws/views.py
```diff
@@ -372,19 +324,23 @@ def htmx_sse_response(response_gen):
             message_html = llm_response_to_html(full_message)
             message_html_lines = message_html.split("\n")
 
+        sync_to_async(llm.create_costs)()
```
juleskuehn, line 327: The query costing isn't working yet. Not sure why.

django/laws/views.py
```diff
@@ -372,19 +324,23 @@ def htmx_sse_response(response_gen):
...
+        sync_to_async(cache.delete)(f"sources_{query}")
+
         yield (
             f"data: <div hx-swap-oob='true' id='answer-sse'>"
             f"<div>{sse_joiner.join(message_html_lines)}</div></div>\n\n"
```
juleskuehn, line 332: Sometimes the AI answer gets streamed out, then disappears and starts over, then completes. Not sure why. This problem may predate this PR.

django/requirements.txt
```diff
@@ -18,6 +18,7 @@ rules==3.4
 django-cleanup==8.1.0
 django-azure-auth==2.0.0
 gevent==24.2.1
+django-data-fetcher==2.2
```
juleskuehn, line 21: Not used in this PR so I could remove, but will need to use it very shortly for another ticket.

smcdonald-jus:
The old form had a few options for searching across enabling authorities that have been removed here (the ones circled below). Are these options worth preserving?

image

juleskuehn:
> The old form had a few options for searching across enabling authorities that have been removed here (the ones circled below). Are these options worth preserving?
> 
> image

Indeed. Nobody has really asked for these, and there is a work around to get those functions so I think it is OK to remove them for now. (For "enabling authority of selected": Just find the document you want with basic search, click kebab icon, you'll see the enabling authority. Then search for that using "select acts/regs"; For "Same enabling authority", again, find the enabling authority you want, then search for that using "enabling authority")

haitmo (approved):

PR #44, opened 2024-09-17
feat: QA document selection; fix: data source selection

juleskuehn:
- [x] implement in chatoptions form
- [x] saving, reset on library change, library modal close handler
- [x] re-implement chunk filtering in QA response view based on new fields
- [x] manually test preset saving/loading
- [x] write new tests

image
image
image


haitmo (approved):
Looks good Jules!

Is there any way we can up the contrast of  values already selected in autocomplete?

image

PR #45, opened 2024-09-18
feat: quick QA file upload (default personal library)

juleskuehn:
* Users each have a "Personal library" which is created by default.
* Chats each have a data source, within the personal library, created by default.
* When files uploaded into QA chat, they are added to the Personal library with Celery task, just like in the librarian modal
* The chat options for QA are updated to select the Personal Library and the Chat's data source.
image


django/librarian/migrations/0007_remove_library_chat_datasource_chat.py
```diff
@@ -0,0 +1,30 @@
...
+
+    operations = [
+        migrations.RemoveField(
+            model_name="library",
+            name="chat",
+        ),
+        migrations.AddField(
+            model_name="datasource",
+            name="chat",
+            field=models.OneToOneField(
```
jannable, line 22: This is OK for now but I generally don't like that DataSource has a chat attribute. Would rather the link be the other way or as a joining table.

django/librarian/models.py
```diff
@@ -88,15 +88,13 @@ class Library(models.Model):
     # Last access time manually updated when library is queried through Library Q&A
     accessed_at = models.DateTimeField(auto_now_add=True)
 
-    chat = models.OneToOneField(
-        "chat.Chat", on_delete=models.CASCADE, null=True, blank=True
-    )
     order = models.IntegerField(default=0)
     is_public = models.BooleanField(default=False)
     is_default_library = models.BooleanField(default=False)
```
jannable, line 93: The distinction between public, default, and personal is becoming less clear. Maybe "type" or "audience" and the options are public/private. And if private, then we rely on the row level access controls. The owner of the object always has permissions. (Again, something we can adjust another time.)

django/librarian/views.py
```diff
@@ -27,7 +27,7 @@
 def get_editable_libraries(user):
     return [
         library
-        for library in Library.objects.filter(chat=None)
```
jannable, line 30: Glad to see the logic improve here.

django/librarian/migrations/0006_alter_library_options_library_is_personal_library.py
```diff
@@ -0,0 +1,25 @@
...
+        migrations.AlterModelOptions(
+            name="library",
+            options={
+                "ordering": ["-is_personal_library", "-is_public", "order", "name"],
+                "verbose_name_plural": "Libraries",
+            },
+        ),
+        migrations.AddField(
+            model_name="library",
+            name="is_personal_library",
```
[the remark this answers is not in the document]
juleskuehn: That is good feedback. I will try changing it to an enum or just enforce a constraint in the validation.

django/librarian/migrations/0007_remove_library_chat_datasource_chat.py
juleskuehn: That makes sense, since every chat will have a data source but not every data source has a chat.
In practice though, doesn't make a difference since you can lookup from the chat side by related name.
Will see about changing it.

django/librarian/models.py
juleskuehn: Thanks, would be nice to brainstorm this. To better explain, "is_default" means the global default for Otto. i.e. all users have this selected initially when they go to Q&A mode. (Corporate library). "is_public" indicates that the library is viewable to all Otto users. There are different permissions and warnings for editing when a library faces all users. (data steward role, etc)

juleskuehn:
I'll address these comments in https://github.com/justicecanada/otto/pull/47 which already has some fixes for this.

PR #47, opened 2024-09-20
feat: agent mode select (Q&A vs. Chat)

juleskuehn:
Merge after https://github.com/justicecanada/otto/pull/45

Fixes a few bugs in that PR.

Makes chat the default mode but with a simple agent (preliminary LLM call) to detect if it should be using Q&A mode instead - and if so, what library to search (based on Library names and descriptions)

image

image

This could be improved in a number of ways, e.g. structured outputs, preliminary RAG over the library/datasource/documents metadata to select the right library, operate in more modes (e.g. switch OUT of Q&A or into another library)


django/chat/models.py
```diff
@@ -14,14 +14,22 @@
...
 
-DEFAULT_MODE = "qa"
+DEFAULT_MODE = "chat"
+
+
+def create_chat_data_source(user):
+    return DataSource.objects.create(
+        name=f"Chat {timezone.now().strftime('%Y-%m-%d %H:%M:%S')}",
+        library=user.personal_library,
+    )
```
juleskuehn, line 32: Refactoring as suggested by Jason to reverse relationship between Chat and DataSource.
Chat now has a data_source property which is populated with this helper when Chat is created (or, for existing chats, when chat is accessed).

django/chat/models.py
```diff
@@ -148,6 +159,7 @@ class ChatOptions(models.Model):
     chat_model = models.CharField(max_length=255, default="gpt-4o")
     chat_temperature = models.FloatField(default=0.1)
     chat_system_prompt = models.TextField(blank=True)
+    chat_agent = models.BooleanField(default=True)
```
juleskuehn, line 162: Agent is on by default in chat mode, to better redirect questions about e.g. HR/pay to the corporate library.
This is heavily dependent on Library names and descriptions (and data source names) being sufficiently descriptive.

django/chat/models.py
```diff
@@ -192,7 +204,7 @@ def qa_prompt_combined(self):
         return ChatPromptTemplate(
             message_templates=[
                 ChatMessage(
-                    content=self.qa_system_prompt,
+                    content=current_time_prompt() + self.qa_system_prompt,
```
juleskuehn, line 207: Most chat tools e.g. ChatGPT, Bing, etc. seem to have this in their system prompt. It helps certain types of queries and only takes a few tokens.

django/chat/prompts.py
```diff
@@ -1,18 +1,20 @@
...
+    "to be better suited for finding relevant and accurate laws and regulations in "
+    "Canada. If relevant, add a markdown link to the Legislation search app.\n\n"
     "3. You do not have access to the internet or other knowledge bases. "
     "If you are asked about very specific facts, especially one about the "
     "Government of Canada or laws, you always caveat your response, e.g., "
     "'I am a pre-trained AI and do not have access to the internet, "
     "so my answers might not be correct. Based on my training data, I expect that...'\n\n"
-    "4. If you are asked a question about Department of Justice or other Government of "
-    "Canada / HR policies, you inform users of Otto's 'Q&A' mode which "
-    "can provide more accurate information.\n\n"
```
juleskuehn, line 14: We no longer need this instruction since the agent will take care of it. (Unless agent is explicitly disabled)

django/chat/responses.py
```diff
@@ -24,44 +26,50 @@
...
     response_message = Message.objects.get(id=message_id)
     chat = response_message.chat
-    assert chat.user_id == request.user.id
     mode = chat.options.mode
 
     # For costing and logging. Contextvars are accessible anytime during the request
     # including in async functions (i.e. htmx_stream) and Celery tasks.
     bind_contextvars(message_id=message_id, feature=mode)
 
+    agent_enabled = not skip_agent and mode == "chat" and chat.options.chat_agent
```
juleskuehn, line 48: We may also want to enable the agent in Q&A mode so that it redirects back to Chat mode when your question is unrelated to the available Q&A libraries.

django/chat/responses.py
```diff
@@ -24,44 +26,50 @@
...
+    if agent_enabled:
+        return chat_agent(chat, response_message)
     if mode == "chat":
-        return chat_response(chat, response_message)
+        return chat_response(chat, response_message, switch_mode=switch_mode)
```
juleskuehn, line 52: At present, it will never switch mode into chat mode, but this is ready if we change our minds.

django/chat/responses.py
```diff
@@ -123,7 +128,6 @@ def stop_response(request, message_id):
     """
     response_message = Message.objects.get(id=message_id)
     chat = response_message.chat
-    assert chat.user_id == request.user.id
```
juleskuehn, line 126: This is handled by decorator access controls now.

django/chat/responses.py
```diff
@@ -448,3 +454,83 @@ def error_response(chat, response_message):
...
+        chat.options.qa_data_sources.clear()
+        chat.options.qa_documents.clear()
+    chat.options.save()
+    llm.create_costs()
+    request = get_request()
+    return otto_response(
+        request,
+        message_id=response_message.id,
+        switch_mode=mode if mode != original_mode else False,
+        skip_agent=True,
```
juleskuehn, line 535: Very important to skip_agent here or else we could end up in an infinite loop!

django/chat/static/chat/js/chatInput.js
```diff
@@ -1,9 +1,6 @@
 const chatPromptMinHeight = 85;
 const chatPromptMaxHeight = 400;
 
-// Set the #chat-action value to the {{mode}}
-document.querySelector('#chat-action').value = '{{ mode }}';
```
juleskuehn, line 5: Cleaning up old code from before ChatOptions saved the mode.

django/chat/utils.py
```diff
@@ -203,14 +208,29 @@ def sse_string(message: str, format=True, dots=False, remove_stop=False) -> str:
...
+                    "chat/components/mode_switch_message.html",
+                    {
+                        "mode": mode,
+                        "mode_str": mode_str,
+                        "library_id": chat.options.qa_library_id,
+                        "library_str": chat.options.qa_library.name,
+                    },
+                )
+                yield sse_string(full_message, format=False, dots=dots_html)
+                await asyncio.sleep(1)
```
juleskuehn, line 224: 1 second delay after switching modes to display the "Switched modes" message to the user.

django/chat/views.py
```diff
@@ -433,6 +430,8 @@ def init_upload(request, chat_id):
     """
     chat = Chat.objects.get(id=chat_id)
     mode = chat.options.mode
+    if mode == "chat":
+        mode = "qa"
```
juleskuehn, line 434: Switch to Q&A mode if dropping a file into Chat mode.

django/librarian/models.py
```diff
@@ -99,25 +99,43 @@ class Meta:
...
-        # Ensure that there is at most 1 default library
-        if (
-            self.is_default_library
-            and Library.objects.filter(is_default_library=True)
-            .exclude(pk=self.pk)
-            .exists()
-        ):
-            raise ValidationError("There can be only one default library")
+        self._validate_default_library()
+        self._validate_personal_library()
```
juleskuehn, line 103: Add some validation rules: personal library can't be the default library, or a public library.

django/otto/fixtures/apps.yaml
```diff
@@ -6,6 +6,14 @@
...
     user_group: AI assistant user
     features:
+      - fields:
+          short_name: "chat"
+          name_en: Chat
+          name_fr: Discuter avec l'IA
+          description_en: Chat with a Justice-hosted AI (like ChatGPT).
+          description_fr: Discuter avec une IA hébergée par Justice (comme ChatGPT).
+          url: chat:chat_with_ai
+          category: ai_assistant
```
juleskuehn, line 16: Move Chat to the top of the list in app tiles - and the first tab in the AI assistant.

django/chat/responses.py
jannable: That seems a little more risky than the other way around.

django/chat/templates/chat/components/chat_input.html
```diff
@@ -51,7 +49,7 @@
                           class="nav-link summarize-option"
                           data-bs-toggle="tooltip"
                           data-bs-placement="top"
-                          data-bs-title='{% trans "Rewrite long text into shorter forms" %}'
+                          data-bs-title='{% trans "Summarize text, files and URLs" %}'
```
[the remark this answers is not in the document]
jannable: I support being able to add a URL into Q&A data source.

jannable (approved):

PR #49, opened 2024-09-23
fix: dont reveal query in GET params; allow back/forward in laws search

juleskuehn:


smcdonald-jus:
Seems to be working pretty well!

One thing I noticed: when you go back/forward on searches, it doesn't retain any changes you may have made to the Advanced search options. Would that be a desirable feature?

juleskuehn:
> Seems to be working pretty well!
> 
> One thing I noticed: when you go back/forward on searches, it doesn't retain any changes you may have made to the Advanced search options. Would that be a desirable feature?

Thanks for checking it out! I agree with you about persisting the advanced search options. I would rather add that to the backlog though.

smcdonald-jus:
I'll go ahead and merge then!

PR #52, opened 2024-09-25
ops: srtm traceability batch 2

jannable:
- Added traceability tags
- Made allowed IPs configurable as variables
- Introduced session configuration settings
- Added abuse monitoring disable prerequisite
- Added network timeouts
- Converted cluster, storage, and db to private
- Added soft delete of storage
- Added backup policy for db
- Alerts for DDoS
- Added gpt-4o-mini
- Variable to enable private networking


setup/.env.sandbox
```diff
@@ -15,6 +15,12 @@ ENTRA_CLIENT_NAME="Otto (Sandbox)"
 # Short abbreviation for the organization to be used in resource names
 ORGANIZATION=JUS
 
+# The IP address range that is allowed to access the application. For multiple IPs, use a comma separator.
+ALLOWED_IPS=199.212.215.11/32
+
+# Flag to enable/disable private network access
+USE_PRIVATE_NETWORK=false
```
smcdonald-jus, line 22: Why not make this True? Presumably that would just mean requiring a VPN connection for access?

jannable: For the sandbox and dev, we won't be able to access the resources if we make them private because Cloud, SSC, and IT Sec can't open ports to our specific workstations.

django/otto/settings.py
```diff
@@ -346,14 +348,18 @@
 
 DEFAULT_AUTO_FIELD = "django.db.models.BigAutoField"
 
+# SC-10: Session Timeout
 SESSION_ENGINE = "django.contrib.sessions.backends.db"
+SESSION_COOKIE_AGE = 1800  # 30 minutes
```
juleskuehn, line 353: Unless there is a policy to require this to be 30 minutes, please leave it at 24 hours as this solves an accessibility requirement. Otherwise we need to display "session timeout" warnings and allow users to extend the session which is a bit of a pain. Can implement this if necessary of course.

PR #54, opened 2024-09-26
feat: source ordering

smcdonald-jus:
Sources can now be ordered by either relevance or reading order. Of course, this won't make a difference for the non-RAG mode.


django/chat/models.py
```diff
@@ -281,7 +288,7 @@ def sorted_files(self):
 
     @property
     def sources(self):
-        return self.answersource_set.all().order_by("-node_score")
+        return self.answersource_set.all()
```
smcdonald-jus, line 291: Removing the ordering here lets their display order be dictated by chat_options.

django/chat/responses.py
```diff
@@ -383,7 +383,7 @@ async def add_files_to_library():
             summary_responses, document_titles
         )
         response_generator = None
-        sources = None
+        source_nodes = None
```
smcdonald-jus, line 386: For the sake of consistency with the per-source stuff below.

django/chat/responses.py
```diff
@@ -417,6 +417,15 @@ async def add_files_to_library():
         input = response_message.parent.text
         source_nodes = retriever.retrieve(input)
 
+        if chat.options.qa_source_order == "reading_order":
+            source_nodes = sorted(
+                source_nodes,
+                key=lambda x: (
+                    x.node.relationships["1"].node_id,
```
smcdonald-jus, line 424: The metadata doesn't contain the document UUID - only the UUID for the data source. I can also envision a situation where multiple docs have the same title or URL. Thus, this seems like the best way to order by Document first (although they won't be alphabetized, of course)

django/chat/templates/chat/modals/advanced_qa_modal.html
```diff
@@ -152,7 +152,7 @@ <h5 class="modal-title" id="advanced-qa-modal-title">
               <em>{% trans "(Required) Must include {context_str}, {query_str}, {pre_instructions}, {post_instructions}." %}</em>
             </p>
             <textarea class="form-control form-control-sm" autocomplete="off" autocorrect="off" autocapitalize="off" spellcheck="false" id="qa_prompt_template-modal" data-inputname="qa_prompt_template" onkeypress="updateQaHiddenField(this);" rows="12"></textarea>
-          </div>
+                  </div>
```
smcdonald-jus, line 155: VS code really hates this thing in particular and moves it an arbitrary distance to the right every time I save. I can't win

django/chat/responses.py
juleskuehn: The metadata does include the document UUID - `node["metadata"]["doc_id"]`. You can retrieve the document itself from that, if you wanted to, i.e. `Document.objects.get(uuid_hex=node["metadata"]["doc_id"])`.

Could you please add a comment in the code about what `node.relationships["1"]` is? (If the doc_id change doesn't get rid of it)

django/chat/templates/chat/modals/advanced_qa_modal.html
juleskuehn: Curious. Do you have DjLint installed in the devcontainer?

django/chat/models.py
```diff
@@ -138,6 +138,11 @@ def from_defaults(self, mode=None, user=None):
     ("summarize", _("Read entire documents (slow, expensive)")),
 ]
 
+qa_source_order_CHOICES = [
```
juleskuehn, line 141: Conventionally within our codebase, this entire name would be uppercase

django/chat/responses.py
smcdonald-jus: Just checked - there's only a `data_source_uuid` attribute in the metadata:
image


Comment added!

django/chat/templates/chat/modals/advanced_qa_modal.html
smcdonald-jus: Why, yes I do

django/chat/responses.py
```diff
@@ -417,6 +417,15 @@ async def add_files_to_library():
...
+                    x.metadata["chunk_number"],
```
juleskuehn, line 425: Trying to remember the sorting code in the Gradio app...
I would have thought that first, it would sort by document relevance (i.e. average or max score of the nodes for each document), then by chunk number.
In any case, I think that is a better way to sort them.

django/librarian/utils/process_engine.py
```diff
@@ -303,9 +303,9 @@ def close_tags(html_string):
     if current_text:
         stuffed_texts.append(current_text)
 
-    for text in stuffed_texts:
+    for i, text in enumerate(stuffed_texts):
         node = TextNode(text=text, id_=str(uuid.uuid4()))
-        node.metadata = metadata
+        node.metadata = dict(metadata, chunk_number=i)
```
juleskuehn, line 308: I haven't seen this syntax before; nice :)

juleskuehn (approved):
*Symbolic approval*, even though I know you will probably push more changes that dismiss this review.

django/chat/models.py
smcdonald-jus: Whoops, it actually was until a find-and-replace that evidently had some unintended consequences

django/chat/templates/chat/modals/advanced_qa_modal.html
smcdonald-jus: Yep, DjLint was to blame. Weird

django/chat/responses.py
smcdonald-jus: Docs are now ordered by max relevance score in each one.
Made a note to try average at some point for comparison, but for now max is easier to implement and fairly intuitive to me.

django/chat/responses.py
smcdonald-jus: Using ref_doc_id now, as discussed. See new sorting logic

jannable (approved):

PR #59, opened 2024-09-27
feat: pruning and granularity in per-source Q&A

smcdonald-jus:


django/chat/models.py
```diff
@@ -289,7 +290,7 @@ def sorted_files(self):
 
     @property
     def sources(self):
-        return self.answersource_set.all()
+        return self.answersource_set.all().order_by("id")
```
smcdonald-jus, line 293: On a seemingly random basis, the bulk creation of AnswerSource objects would reorder the sources. I assume this has something to do with the async'ing, because the Django docs say that bulk_create should retain the same order as the input list. This fix ensures that the order established in qa_response is same as their display order in the modal.

django/chat/models.py
```diff
@@ -331,6 +332,7 @@ class AnswerSource(models.Model):
     node_score = models.FloatField(default=0.0)
     # Saved citation for cases where the source Document is deleted later
     saved_citation = models.TextField(blank=True)
+    group_number = models.IntegerField(default=0)
```
smcdonald-jus, line 335: This is used to group sources when displaying them in the modal, for granular Q+A

django/chat/responses.py
```diff
@@ -457,18 +435,82 @@ async def add_files_to_library():
                 content_type="text/event-stream",
             )
 
+        # If we're stitching sources together into groups...
+        if chat.options.qa_granularity > 768:
```
smcdonald-jus, line 439: Note that the granularity resets to 768 in "combine sources" mode, so this block gets bypassed in that case

django/chat/models.py
juleskuehn: Well, it does retain the same order (of IDs) :)

django/chat/templates/chat/modals/advanced_qa_modal.html
```diff
@@ -1,161 +1,262 @@
 {% load i18n %}
 
-<div class="modal modal-xl fade"
-     id="advanced-qa-modal"
-     tabindex="-1"
-     aria-labelledby="advanced-qa-modal-title"
-     aria-hidden="true">
+<div
+  class="modal modal-xl fade"
```
juleskuehn, line 4: Formatter issues. Please ensure djLint is installed and enabled, then save the HTML files again

django/chat/models.py
smcdonald-jus: That still appears to be inconsistent with the stated behaviour of bulk_create. Django docs suggest that the resulting objects will retain the order of the input list, but after running this line in utils.save_sources_and_update_security_label:

`AnswerSource.objects.bulk_create(sources)`

I'd randomly see sources (the input list) ordered differently from the resulting message.sources (the created model instances). This stopped happening when I introduced the explicit ordering above. Aside from a possible error in the Django source code, async'ing is the only thing I can think of that'd cause this - and it does appear to be "genuinely" random.

PR #66, opened 2024-10-08
fix (laws): use instructions in AI answer

juleskuehn:


django/laws/prompts.py
```diff
@@ -21,9 +21,9 @@
     "---------------------\n"
     "Given the context information and not prior knowledge, "
     "answer the query.\n"
-    "If the context information is entirely unrelated to the provided query, "
-    "don't try to answer the question; just say 'Sorry, I cannot answer "
-    "that question.'.\n"
+    # "If the context information is entirely unrelated to the provided query, "
+    # "don't try to answer the question; just say 'Sorry, I cannot answer "
+    # "that question.'.\n"
```
juleskuehn, line 26: This instruction was interfering with GPT-4o-mini ability to follow the user-given instructions. Left as a comment to make its removal obvious, in case an issue emerges.

smcdonald-jus (approved):

PR #67, opened 2024-10-09
feat: force OCR button

juleskuehn:
Plus fix some performance issues & features to enable pilot use cases.


django/chat/models.py
```diff
@@ -89,6 +89,11 @@ def access(self):
         self.accessed_at = timezone.now()
         self.save()
 
+    def delete(self, *args, **kwargs):
```
smcdonald-jus, line 92: Doesn't Django already automatically delete related model instances? (you might need Cascade mode for it, iirc)?
Or is the idea that we want to keep _some_ of those objects (i.e. Costs)?

smcdonald-jus (approved):

PR #73, opened 2024-10-15
fix: use advanced instructions in per-doc mode

smcdonald-jus:


juleskuehn (approved):
Good! I think this makes sense to do. Tried it briefly and the results look good (not surprisingly, mostly the same as on RAG mode)

Deleting the prompts from the Advanced Q&A Options modal reverts the behaviour so that's still an option.

PR #74, opened 2024-10-15
feat: user cost display and limits

juleskuehn:
Looking for feedback on both the UI and implementation. 

image
image
image

Updates every 60 seconds and after each chat response.

This PR does not address actually imposing the limits, only alerting the user to their usage.

There will be at least 1 follow-up PR to do the following:
- ~block user's access when limit exceeded (decorator on cost-initiating views?)~ **done in this PR**
- ~ability to override weekly limit temporarily for a user (otto admins - manage users page)~ **done in this PR**
- fix the French numbers bug in the cost management dashboard - **larger problem actually; opened new ticket**


django/otto/templates/components/user_cost.html
```diff
@@ -0,0 +1,22 @@
+<div id="cost-bar-outer"
+     hx-get="{% url 'user_cost' %}"
+     hx-swap="outerHTML"
+     hx-trigger='load delay:60000ms'
```
haitmo, line 4: Is there a need to reload costs every 60 seconds?

django/otto/templates/components/user_cost.html
```diff
@@ -0,0 +1,22 @@
...
+     data-bs-title="{{ cost_tooltip }}"
+     style="position: absolute;
+            bottom: 0px;
+            width: 100%;
+            height: 5px;
+            outline: solid 1px #666;
+            z-index: 1;
+            border-radius: 3px;
+            overflow: hidden">
+     <div id="cost-bar-inner"
```
haitmo, line 17: Would using the meter html element be helpful here to simplify some of the logic / closer semantically to what we are trying to achieve?

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meter

haitmo (approved):
Looks Good Jules!

As a quick solution to the problem I think this is great.

I do think this is the wrong place for it though. Personally I would rather keep this information in a "User" menu and have alert triggers when some thresholds are crossed or when money is "spent"

A user menu would require a bit more planning, I'd be happy to sit with you and design it together. Not urgent at all though.

django/otto/templates/components/user_cost.html
juleskuehn: Wasn't aware of this. That could work if the color order is reversed with CSS so low = green, high = red.

django/otto/templates/components/user_cost.html
juleskuehn: Kind of? I am OK with disabling the polling. I thought you might have an opinion on this.

Rationale for polling: not everything that incurs costs will automatically update the cost widget.
Specifically, actions in the Librarian modal (such as uploading/re-processing documents) don't update the cost widget. 
 There is also the possibility someone has multiple Otto tabs open. And then there are apps other than the AI assistant which can incur costs.

haitmo: No all good, thanks for the info

django/otto/views.py
```diff
@@ -654,3 +662,25 @@ def cost_dashboard(request):
...
+
+def user_cost(request):
+    today_cost = cad_cost(Cost.objects.get_user_cost_today(request.user))
+    weekly_max = request.user.this_week_max
+    this_week_cost = cad_cost(Cost.objects.get_user_cost_this_week(request.user))
+    cost_percent = max(
+        min(int(100 * this_week_cost / weekly_max if weekly_max else 0), 100), 1
+    )
+    cost_tooltip = (
+        f"${this_week_cost:.2f} / ${weekly_max:.2f} {_('this week')}<br>"
```
haitmo, line 675: Small thing but in French, the $ sign goes after the number. Not sure how this is handled in python. In javascscript I usually stick to Intl.NumberFormat

haitmo (changes requested):
Looks good! There is an issue with the app however:

Something is causing an infinite loop and continuously making a request to the user_cost view if the terms of use haven't been accepted.

I think load_app_localization needs to be run as well

haitmo (approved):

PR #75, opened 2024-10-16
fix: failed translation failure

haitmo:
Fixes 500 error when trying to download resulting translated file.


juleskuehn (approved):
Thanks for catching this!

PR #78, opened 2024-10-16
feat: extract page numbers (PDF, PPTX); refactor: parsing/chunking

Rifahaziz:


django/librarian/tasks.py
```diff
@@ -117,7 +117,10 @@ def process_document_helper(document, llm, force_azure=False):
                 "status_text": _("Adding to library..."),
             },
         )
+    print("---------------------Extracted text----------------")
+    print(document.extracted_text)
```
smcdonald-jus, line 121: Does printing document content to the logs have any implications in terms of data security, retention, etc.? Obviously there isn't really any increased risk, but I'm wondering if it might have "on-paper" implications for compliance.

juleskuehn: Will be removed in next push. Thx!

juleskuehn (approved):

PR #79, opened 2024-10-17
perf: limit redundant files

juleskuehn:
This addresses ballooning storage issues with redundant file uploads and some technical debt re: unused files not being deleted. **Also adds more cleanup commands to Celery Beat, and tests them. Plus a few unrelated fixes to manage users CSV upload/download.**

* reuse SavedFile objects if possible to avoid duplicate files on disk
* delete SavedFile when there are no Document or ChatFile objects referencing it
* check hash client-side when file is uploaded in chat; if it exists on server, don't upload at all.

To test functionality:
* Delete all your chats and empty your media directory.
* Start a new chat.
* Drag a file in.
* Check the media directory for the file
* Drag the same file in (even with a different filename will work)
* Check the media directory - there should still be only 1 file.
* Both files should be in the library.
* Delete the chat
* Check the media directory - files should be gone.

You can also test from the Librarian modal which works similarly:
* Upload a file through Librarian modal.
* Check media directory
* Upload same file again.
* Check media directory - still should be just 1 instance of the file though there are now 2 documents
* Delete one of the documents in Librarian modal.
* File should still be there, since there is a reference from the other Document object.
* Delete the other Document
* File should be gone now.


django/librarian/models.py
```diff
@@ -411,6 +411,11 @@ def delete(self, *args, **kwargs):
             self.remove_from_vector_store()
         except Exception as e:
             logger.error(f"Failed to remove document from vector store: {e}")
+        file = self.file
+        if file:
+            self.file = None
+            self.save()
+            file.safe_delete()
```
juleskuehn, line 418: This could probably also be accomplished via the post_delete handler but this seems to be working. I could / should add tests.

django/librarian/models.py
```diff
@@ -458,8 +463,17 @@ def __str__(self):
...
         if self.file:
             with self.file.open("rb") as f:
-                sha256 = hashlib.sha256(f.read())
-                self.sha256_hash = sha256.hexdigest()
+                self.sha256_hash = generate_hash(f.read())
                 self.save()
+
+    def safe_delete(self):
+        if self.chat_files.exists() or self.documents.exists():
+            logger.info(f"File {self.file.name} has associated objects; not deleting")
```
juleskuehn, line 475: Basically, garbage collection of objects with no references.

django/librarian/views.py
```diff
@@ -470,8 +471,18 @@ def upload(request, data_source_id):
...
+        file_exists = SavedFile.objects.filter(sha256_hash=file_hash).exists()
+        if file_exists:
+            file_obj = SavedFile.objects.filter(sha256_hash=file_hash).first()
+            logger.info(
+                f"Found existing SavedFile for {file.name}", saved_file_id=file_obj.id
+            )
+        else:
+            file_obj = SavedFile.objects.create(content_type=file.content_type)
+            file_obj.file.save(file.name, file)
+            file_obj.generate_hash()
```
juleskuehn, line 485: The basic librarian file upload needs to also be replaced with JS version, like in the Chat upload above, but I will do that in another PR.

jannable (dismissed):

django/otto/celery.py
```diff
@@ -32,6 +32,21 @@
...
+        "task": "otto.tasks.delete_old_chats",
+        "schedule": crontab(hour=2, minute=0),
+    },
+    # Delete empty chats every day at 2 am UTC
+    "delete-empty-chats-every-morning": {
+        "task": "otto.tasks.delete_empty_chats",
+        "schedule": crontab(hour=2, minute=0),
+    },
+    # Delete unused libraries every day at 3 am UTC
+    # "delete-unused-libraries-every-morning": {
```
juleskuehn, line 46: I am not sure this is a good idea, but I've added it as a comment so that it is clear we have decided not to do this.

jannable (approved):

PR #80, opened 2024-10-18
fix: dragging file into chat / uploading file clears the prompt

haitmo:
Not sure if its the most elegant way of doing it, but it seems to work.


juleskuehn (approved):
Can't think of a reason this would NOT work. Good enough :)

PR #82, opened 2024-10-18
fix: librarian misc errors

juleskuehn:
* Improves document status naming
* Data source "stop" button now only affects incomplete files
* Librarian close modal handler also stops polling Document statuses
* Librarian close modal handler doesn't reset search mode, data source etc. unless library changed
* Uploading files as first message to Q&A clears the placeholders (this bug was introduced in the "don't clear prompt" PR)
* Uploading many files to chat would truncate the message without displaying "show all" button (fixed with timeout)


juleskuehn:
I am seeing some of these pop up in Django or Celery logs. Will to try address in this PR also.
`-> Cannot close object, library is destroyed. This may cause a memory leak!`

django/chat/static/chat/js/scripts.js
```diff
@@ -137,11 +137,11 @@ document.addEventListener("DOMContentLoaded", function () {
 // On prompt form submit...
 document.addEventListener("htmx:afterSwap", function (event) {
   if (event.target.id != "messages-container") return;
-  if (event.detail.pathInfo.requestPath.includes('upload'))
-    return;
   if (document.querySelector("#no-messages-placeholder") !== null) {
     document.querySelector("#no-messages-placeholder").remove();
   }
+  if (event.detail.pathInfo.requestPath.includes('upload'))
+    return;
```
juleskuehn, line 144: Needed to move this down here to ensure that the placeholder is removed when user uploads a file to a new chat.

django/chat/templates/chat/components/chat_message.html
```diff
@@ -63,10 +63,14 @@
...
-  if (element && (element.offsetHeight < element.scrollHeight)) {
-    element.closest('.message-outer').classList.add('truncate');
+  function checkTruncation() {
+    let element = document.querySelector('#message_{{message.id}} .message-text');
+    if (element && (element.offsetHeight < element.scrollHeight)) {
+      element.closest('.message-outer').classList.add('truncate');
+    }
   }
+  checkTruncation();
+  setTimeout(checkTruncation, 100);
```
juleskuehn, line 73: Needed to run this code again after a slight delay due to the message needing time to render before its height can be measured accurately. (This is to display the "show more" button when message is truncated)

django/librarian/static/librarian/scripts.js
```diff
@@ -10,11 +10,18 @@ let librarianModalCloseHandler = null;
...
   const modalEl = document.getElementById('editLibrariesModal');
-  const chatId = modalEl.getAttribute('data-chat-id');
   modalEl.removeEventListener('hidden.bs.modal', librarianModalCloseHandler);
   librarianModalCloseHandler = event => {
+    // Stop polling on element id="libraryModalPoller"
+    // by replacing it with empty div
+    const poller = document.getElementById('libraryModalPoller');
+    const newPoller = document.createElement('div');
+    newPoller.id = 'libraryModalPoller';
+    poller.parentNode.replaceChild(newPoller, poller);
```
juleskuehn, line 20: When the librarian modal is closed, we want it to stop polling for the document status updates. This slows the page down unnecessary since user is not looking at the status changes anymore.

django/librarian/static/librarian/scripts.js
```diff
@@ -10,11 +10,18 @@ let librarianModalCloseHandler = null;
...
+    // Update the QA library select
     const qa_library_select = document.getElementById("id_qa_library");
-    if (qa_library_select.value !== library_id) {
+    // Only update if the value is different (allow type coercion)
+    if (qa_library_select.value != library_id) {
```
juleskuehn, line 24: Needed to replace `!==` with `!=` to allow type coersion because one of these is a string and one is an int.

django/librarian/views.py
```diff
@@ -430,7 +430,8 @@ def data_source_stop(request, data_source_id):
     # Stop all celery tasks for documents within this data source
     data_source = get_object_or_404(DataSource, id=data_source_id)
     for document in data_source.documents.all():
-        document.stop()
+        if document.status in ["PENDING", "INIT", "PROCESSING"]:
+            document.stop()
```
juleskuehn, line 434: The "stop" button should really only stop documents that are currently being processed.

django/tests/text_extractor/helpers/test_views.py
```diff
@@ -73,9 +73,9 @@ def test_document_submission_and_download(client, all_apps_user, mock_pdf_file):
     docs = response.context["ocr_docs"]
     assert len(docs) == 1
 
-    # There should be two properties on the doc: "pdf" and "txt". Each has a file_id
-    pdf_file_id = docs[0]["pdf"]["file"].file_id
-    txt_file_id = docs[0]["txt"]["file"].file_id
+    # There should be two properties on the doc: "pdf" and "txt". Each has an ID.
+    pdf_file_id = docs[0]["pdf"]["file"].id
+    txt_file_id = docs[0]["txt"]["file"].id
```
juleskuehn, line 78: In a recently merged PR, the "file_id" field was removed from the object.

django/tests/text_extractor/helpers/test_views.py
```diff
@@ -98,9 +98,10 @@ def test_document_submission_and_download(client, all_apps_user, mock_pdf_file):
     assert "attachment" in response["Content-Disposition"].lower()
 
     # Try a random File ID; this should return the error message ("error" in the response)
+    import uuid
+
+    random_id = uuid.uuid4()
```
juleskuehn, line 103: The OutputFile class inherits from SecureModel, which has UUID for IDs.

smcdonald-jus (approved):

django/tests/text_extractor/helpers/test_views.py
Rifahaziz: tests working fine now.

django/librarian/templates/librarian/components/document_status.html
Rifahaziz, line 1: works well!

Rifahaziz (approved):

PR #84, opened 2024-10-23
fix: long filenames handled in text-extractor

Rifahaziz:


mcusteau (approved):

django/text_extractor/utils.py
```diff
@@ -285,3 +285,12 @@ def create_searchable_pdf(input_file, add_header):
         output.add_page(new_pdf_page.pages[0])
 
     return output, all_text
+
+
+def shorten_input_name(input_name, max_length=35):
```
juleskuehn, line 290: You may as well just always use a UUID, then you never have to try to make filenames unique, right?

django/text_extractor/utils.py
```diff
@@ -285,3 +285,12 @@ def create_searchable_pdf(input_file, add_header):
...
+    base_name, file_extension = os.path.splitext(input_name)
+    if len(base_name) > max_length:
+        unique_id = str(uuid.uuid4())
+        shortened_base_name = base_name[:10] + "_" + unique_id
```
juleskuehn, line 294: You can just use the UUID, no need for the base name (although... it may be nice for debugging. Your call).

juleskuehn (commented):
This looks like it will solve the problem!
I think it can be made simpler so that you only keep track of the original filename in the ContentFile and OutputFile objects, and can just use the UUID filenames everywhere else.
Let's call to discuss.

PR #85, opened 2024-10-24
fix: text extractor filename length

juleskuehn:
Based off Rifah's other open PR


django/text_extractor/models.py
```diff
@@ -13,9 +13,8 @@ def __str__(self):
 
 
 class OutputFile(SecureModel):
-    file_id = models.CharField(max_length=255, null=True, blank=True)
     file = models.FileField(upload_to="ocr_output_files/")
-    file_name = models.CharField(max_length=255, null=True, blank=True)
+    file_name = models.TextField(default="tmp")
```
juleskuehn, line 17: any input filename will work now (in this field)

django/text_extractor/templates/text_extractor/completed_documents.html
```diff
@@ -4,15 +4,13 @@
   <tr>
     <td>
       <i class="bi bi-filetype-pdf me-1 h5" style="color: red"></i>
-      <a href="{% url 'text_extractor:download_document' doc.pdf.file.file_id user_request_id %}"
-         class="link-dark"
-         target="_blank">{{ doc.pdf.file.file_name }} (PDF, {{ doc.pdf.size }})</a>
+      <a href="{% url 'text_extractor:download_document' doc.pdf.file.id user_request_id %}"
+         class="link-dark">{{ doc.pdf.file.file_name }} (PDF, {{ doc.pdf.size }})</a>
```
juleskuehn, line 8: don't need _blank since it will download a file anyway

django/text_extractor/utils.py
```diff
@@ -285,3 +285,8 @@ def create_searchable_pdf(input_file, add_header):
         output.add_page(new_pdf_page.pages[0])
 
     return output, all_text
+
+
+def shorten_input_name(input_name):
+    base_name, file_extension = os.path.splitext(input_name)
+    return str(uuid.uuid4()) + file_extension
```
juleskuehn, line 292: Use a UUID for file to ensure uniqueness and length (no need to prepend username and date etc.)

django/text_extractor/views.py
```diff
@@ -88,26 +87,26 @@ def submit_document(request):
                     if idx > 0:  # Exclude TOC from file names to merge
                         file_names_to_merge.append(file_name)
                 else:
-                    file_id = f"{user_name}_{current_time}_OCR_{input_name}.pdf"
-                    text_id = f"{user_name}_{current_time}_{input_name}.txt"
```
juleskuehn, line 92: Don't need these special IDs, the normal UUID provided by SecureModel will be fine

django/text_extractor/views.py
```diff
@@ -88,26 +87,26 @@ def submit_document(request):
...
-                    file_name = f"OCR_{input_name}.pdf"
-                    text_name = f"OCR_{input_name}.txt"
+                    file_name = f"{input_name}_OCR.pdf"
+                    text_name = f"{input_name}_OCR.txt"
 
-                    content_file = ContentFile(pdf_bytes.getvalue(), name=file_id)
-                    content_text = ContentFile(txt_file, name=text_id)
+                    content_file = ContentFile(
+                        pdf_bytes.getvalue(), name=shorten_input_name(file_name)
```
juleskuehn, line 94: We use the new generated (short, unique) filename only here, for the file that is actually written to /media folder.

django/text_extractor/views.py
```diff
@@ -213,9 +206,7 @@ def download_document(request, file_id, user_request_id):
     user_request = UserRequest.objects.get(access_key, id=user_request_id)
 
     try:
-        output_file = user_request.output_files.get(
-            access_key=access_key, file_id=file_id
-        )
+        output_file = user_request.output_files.get(access_key=access_key, id=file_id)
```
juleskuehn, line 209: Since there is no special "file_id" field anymore, we just use the normal ID inherited from SecureModel

mcusteau (approved):

PR #88, opened 2024-10-28
fix: text extractor files are deleted after 24 hrs

Rifahaziz:
Files older than 24 hours deleted, tests written.


django/otto/celery.py
```diff
@@ -47,6 +47,10 @@
     #     "task": "otto.tasks.delete_unused_libraries",
     #     "schedule": crontab(hour=3, minute=0),
     # },
+    "delete-text-extractor-files-every-day": {
+        "task": "otto.tasks.delete_text_extractor_files",
+        "schedule": crontab(hour=0, minute=0),
+    },
 }
```
Rifahaziz, line 55: task added to celery to delete after every 24 hrs

django/tests/otto/test_cleanup.py
```diff
@@ -300,3 +305,91 @@ def test_delete_empty_chats_task(client, all_apps_user):
...
+    new_output_files = OutputFile.objects.filter(
+        access_key=access_key, file_name="test_file2.txt"
+    )
+    assert new_output_files.count() == 1
+
+    # Check media directory
+    media_folder = os.path.join(settings.MEDIA_ROOT, "ocr_output_files")
+    print(f"Files in media folder: {os.listdir(media_folder)}")
+    assert "test_file1.txt" not in os.listdir(media_folder)
+    assert "test_file2.txt" in os.listdir(media_folder)
```
Rifahaziz, line 395: 2 test files created, one is 40 hours old and the second is 5 minutes old.
checks:
-First file should be deleted as it is 40 hours old
-Second file should be kept as it is 5 minutes old

django/text_extractor/models.py
```diff
@@ -18,6 +18,7 @@ class OutputFile(SecureModel):
     user_request = models.ForeignKey(
         UserRequest, related_name="output_files", on_delete=models.CASCADE
     )
+    created_at = models.DateTimeField(auto_now_add=True)
```
Rifahaziz, line 21: Previously it was using UserRequest model's creation time which messed up the times. Added creation time to OutputFile model now.

django/otto/management/commands/delete_text_extractor_files.py
```diff
@@ -0,0 +1,34 @@
...
+        for output_file in old_files:
+            file_path = output_file.file.path
+            if os.path.exists(file_path):
+                os.remove(file_path)
+                self.stdout.write(
+                    self.style.SUCCESS(
+                        f"Deleted file: {file_path}, created_at: {output_file.created_at}"
+                    )
+                )
+            output_file.delete(access_key=access_key)
```
Rifahaziz, line 34: filters old files based on the cutoff time (older than 24 hrs)

django/tests/otto/test_cleanup.py
Rifahaziz: Permissions had to be explicitly given to create OutputFile objects

django/text_extractor/models.py
juleskuehn: Isn't it OK to use UserRequest creation time? They should only differ by at most like 1 minute right?
See my comment above about, "why not delete UserRequest".

juleskuehn (commented):
Hey Rifah, this mostly looks right and does meet the requirement that OutputFiles are cleaned up from disk.
But, it can be made simpler - plus clean up more of the database objects - by making a few changes.

If you delete the UserRequest instead of the OutputFiles, this will cascade delete OutputFiles as well. The file itself should be deleted from disk automatically by the "django_cleanup" app (what you did - the file from disk manually - is usually correct, but I added this app to make it easier).

Do these suggestions for changes sound reasonable? Please let me know if I'm missing something.
* remove the code for manually deleting files
  * check that tests still pass
* delete UserRequest objects instead of OutputFile
  * modify test so that UserRequest created_at is modified instead of OutputFile (which no longer exists)
  * check that tests pass
* add test that UserRequest is deleted too
  * check that tests pass
* undo changes to OutputFile model
  * run makemigrations and migrate on your own computer
  * delete the migration files from this PR (after you've )
  * check that tests pass

django/text_extractor/models.py
Rifahaziz: Deleting userRequest works fine, but in the test it is not automatically deleting the files as it is doing normally with the django cleanup. Any suggestions on how to use the django cleanup on tests?

juleskuehn (dismissed):
Great! If tests pass, looks good to me. I'll run the tests locally as well.

django/tests/otto/test_cleanup.py
```diff
@@ -300,3 +305,120 @@ def test_delete_empty_chats_task(client, all_apps_user):
     # Check that the too-new empty chat remains
     too_new_empty_chat = Chat.objects.filter(id=too_new_empty_chat_id).first()
     assert too_new_empty_chat is not None
+
+
+@pytest.mark.django_db(transaction=True)
+def test_delete_text_extractor_files_task(client, all_apps_user):
+
+    # Ensure the "Otto admin" group exists
+    group, created = Group.objects.get_or_create(name="Otto admin")
```
Rifahaziz, line 314: The "Otto admin" group needed to be manually created to add the permissions. This is why the test was failing.

juleskuehn (approved):

PR #92, opened 2024-10-30
feat: text extractor costs added

Rifahaziz:
Costs are displayed to the user for every output and saved in the cost dashboard.


django/text_extractor/utils.py
```diff
@@ -284,7 +286,24 @@ def create_searchable_pdf(input_file, add_header):
...
-    return output, all_text
+    # Calculate token counts
+    input_token_count = len(all_text.split())
+    output_token_count = len(all_text.split())
+    embed_token_count = 0  # Assuming no embedding tokens for this example
+
+    # Create cost objects
+    usd_cost = 0
+    if input_token_count > 0:
+        c1 = Cost.objects.new(cost_type="doc-ai-read", count=input_token_count)
```
Rifahaziz, line 297: costs based on texts' tokens

django/text_extractor/views.py
```diff
@@ -179,10 +184,11 @@ def submit_document(request):
                         },
                     }
                 )
-
+            formatted_total_cost = display_cad_cost(total_cost)
```
Rifahaziz, line 187: formatting in CAD

django/text_extractor/views.py
```diff
@@ -70,10 +71,14 @@ def submit_document(request):
                 )
                 files.insert(0, toc_file)
 
-            current_time = datetime.now().strftime("%Y%m%d_%H%M%S")
-
             for idx, file in enumerate(files):
-                ocr_file, txt_file = create_searchable_pdf(file, merged and idx > 0)
+                bind_contextvars(
```
Rifahaziz, line 75: saving the cost in dashboard

django/text_extractor/utils.py
juleskuehn: The document intelligence API is costed by number of pages. You only need to create the 1 cost object:

```
num_pages = # Number of pages you sent to the API
cost = Cost.objects.new(cost_type="doc-ai-read", count=num_pages)
```

There is an example of this in `process_engine.py`

(There are no tokens (input, output or embedding) being used because you are not calling an LLM).

django/text_extractor/views.py
juleskuehn: Good - you can just put this at the top of the submit_document function though, rather than in this loop. It just needs to be set once.

django/text_extractor/views.py
```diff
@@ -179,10 +184,11 @@ def submit_document(request):
                         },
                     }
                 )
-
+            formatted_total_cost = display_cad_cost(total_cost)
             context = {
                 "ocr_docs": completed_documents,
                 "user_request_id": user_request.id,
+                "total_cost": formatted_total_cost,
```
juleskuehn, line 191: You should be passing through the cost in the ocr_docs dict (since cost is for each ocr_doc in the table).
Currently, if you processed 5 docs (and merge is off) and the costs were $0.01, $0.02, $0.03, $0.04, $0.05, there would be 5 rows output for the 5 ocr docs, right?
But since you are only passing 1 cost, it will say that each document cost $0.15.

juleskuehn (commented):
Needs a few changes - see comments.

Can you also write a test to ensure that a cost object is correctly created for "text extractor"? I don't think there are currently any tests for the views. I can help you with this if you want. It should be similar to the tests for librarian.

django/text_extractor/views.py
Rifahaziz: good catch! fixing it.

django/text_extractor/utils.py
```diff
@@ -286,22 +286,11 @@ def create_searchable_pdf(input_file, add_header):
...
-    if embed_token_count > 0:
-        c3 = Cost.objects.new(cost_type="embedding", count=embed_token_count)
-        usd_cost += c3.usd_cost
+    num_pages = len(ocr_results.pages)
+
+    # Create a single cost object based on the number of pages
+    cost = Cost.objects.new(cost_type="doc-ai-read", count=num_pages)
+    usd_cost = cost.usd_cost
 
     return output, all_text, usd_cost
```
Rifahaziz, line 296: one cost based on page numbers added

django/text_extractor/views.py
```diff
@@ -184,12 +184,17 @@ def submit_document(request):
...
+                context = {
+                    "ocr_docs": completed_documents,
+                    "user_request_id": user_request.id,
+                    "total_cost": formatted_total_cost,
+                }
+            else:
+                context = {
+                    "ocr_docs": completed_documents,
+                    "user_request_id": user_request.id,
+                }
```
Rifahaziz, line 197: single costs associated when files are not merged, otherwise total cost is passed.

Rifahaziz:
> Needs a few changes - see comments.
> 
> Can you also write a test to ensure that a cost object is correctly created for "text extractor"? I don't think there are currently any tests for the views. I can help you with this if you want. It should be similar to the tests for librarian.

All changes made that you asked for, except adding the tests.

juleskuehn (approved):

PR #93, opened 2024-10-30
feat: librarian bulk actions and Azure Read API

juleskuehn:
- [x] "Process all" or "Force OCR all" to a single "Bulk actions" dropdown with different options, e.g. "Process incomplete", "Process all" (with and without OCR options)
- [x] Track what method was used to process the document (OCR or not) and display in document status pane
- [x] Add Azure Document Intelligence Read API (cheaper than Layout, but still has OCR)
- [x] Skip markdown processing (e.g. don't try to repeat headings) in splitter when appropriate (e.g. from non-markdown plain text file, or non-layout PDF extraction)
- [x] Fix some small bugs in document loading, e.g. extra newlines.
- [x] Add page range and page crop options to model
  * **TODO** implement in the document loaders - maybe not in this PR
- [x] Fix error -> Cannot close object, library is destroyed. This may cause a memory leak!
- [x] Upload limit: data source <= 100 files
- [x] Upload limit: total file size <= 300 mb

image
image
image
image


django/chat/models.py
```diff
@@ -411,7 +411,7 @@ class ChatFile(models.Model):
     def __str__(self):
         return f"File {self.id}: {self.filename}"
 
-    def extract_text(self, fast=True):
+    def extract_text(self, pdf_method="default"):
```
juleskuehn, line 414: Instead of just "fast" (local with pdfium) and "not fast" (azure prebuilt-layout), we can supply a string for pdf_method.

Now, these will be one of:
* default (pdfium)
* azure_read
* azure_layout

We can easily add or change methods in the future.

django/chat/views.py
```diff
@@ -366,26 +366,29 @@ def chunk_upload(request, message_id):
...
     file = request.FILES["file"].read()
     content_type = request.POST["content_type"]
-    fileName = request.POST["filename"]
+    file_name = request.POST["filename"]
     file_id = request.POST["file_id"]
     end = request.POST["end"]
     nextSlice = request.POST["nextSlice"]
 
-    if file == "" or fileName == "" or file_id == "" or end == "" or nextSlice == "":
+    if ("text" in content_type or not content_type) and file_name.endswith(".md"):
```
juleskuehn, line 374: Markdown content type is not usually detected. We will just assume it is markdown if file is plain text and ends in "md".

django/librarian/models.py
```diff
@@ -344,6 +348,22 @@ class Meta:
...
+    # Page range to extract from PDF (e.g. "1-3, 5, 7-9")
+    pdf_page_range = models.CharField(max_length=255, null=True, blank=True)
+    # Crop box for PDF extraction (left, right, top, bottom) - ratio of page size
+    pdf_crop_left = models.FloatField(null=True, blank=True)
+    pdf_crop_right = models.FloatField(null=True, blank=True)
+    pdf_crop_top = models.FloatField(null=True, blank=True)
+    pdf_crop_bottom = models.FloatField(null=True, blank=True)
+    # The number of detected / extracted pages in the PDF (for debugging / confirmation)
+    pdf_num_pages_detected = models.IntegerField(null=True, blank=True)
+    pdf_num_pages_extracted = models.IntegerField(null=True, blank=True)
```
juleskuehn, line 365: We aren't using all the fields in the model right now. I am pushing off the work on PDF crop and page ranges to a later PR.

django/librarian/utils/markdown_splitter.py
```diff
@@ -7,12 +7,15 @@
...
-    def __init__(self, chunk_size=768, chunk_overlap=100, debug=False):
+    def __init__(
+        self, chunk_size=768, chunk_overlap=100, enable_markdown=True, debug=False
+    ):
         self.chunk_size = chunk_size
         self.chunk_overlap = chunk_overlap
         self.debug = debug
         self.current_headings = {i: None for i in range(1, 7)}
         self.last_table_header = None
+        self.enable_markdown = enable_markdown
```
juleskuehn, line 18: Users may upload plain text files or PDFs that have markdown-looking content (lines starting with #, etc) but we know that they aren't actual markdown. In those cases we don't want to do any special processing based on markdown syntax.

django/librarian/utils/process_engine.py
```diff
@@ -180,22 +200,24 @@ def extract_markdown(
...
     text = ""
+    pdf = pdfium.PdfDocument(content)
     for i, page in enumerate(pdf):
+        text_page = page.get_textpage()
         text += f"<page_{i+1}>\n"
-        text += page.get_textpage().get_text_range() + "\n"
+        text += text_page.get_text_range() + "\n"
         text += f"</page_{i+1}>\n"
+        text_page.close()
+    pdf.close()
```
juleskuehn, line 220: The explicit `.close()` is necessary with pypdfium to avoid memory leaks (or at least, warnings about memory leaks).

django/librarian/utils/process_engine.py
```diff
@@ -459,3 +481,47 @@ def _pdf_to_html_using_azure(content):
...
+        for line in page.lines:
+            chunk = {
+                "page_number": page.page_number,
+                "text": line.content + "\n",
+            }
+            p_chunks.append(chunk)
+
+    text = ""
+    cur_page = None
+    for _, chunk in enumerate(p_chunks, 1):
```
juleskuehn, line 514: I'm *not* sorting these by X,Y coordinates. I found that actually made it worse.

django/librarian/utils/markdown_splitter.py
Rifahaziz: very good point

django/librarian/templates/librarian/components/document_list.html
```diff
@@ -31,26 +32,83 @@ <h6 id="documents-label">{% trans "Documents" %}</h6>
...
+                <span>{% trans "PDF layout & OCR" %}</span>
+                <span class="text-muted">{% trans "(best, $$$)" %}</span>
+              </a>
+            </li>
+            <li>
+              <hr class="dropdown-divider">
+            </li>
+            <li>
+              <h6 class="dropdown-header">{% trans "Process incomplete only" %}</h6>
+            </li>
```
Rifahaziz, line 74: Not very clear what "process incomplete only" means, I'm guessing it means it will process the files that have failed before?

Rifahaziz (approved):

PR #94, opened 2024-10-30
feat: conditional display of Q+A options

smcdonald-jus:
For better clarity, only display relevant options in Advanced Q&A. All options that only pertain to RAG are hidden in whole-doc-summary mode, and the toggle for pruning is hidden in "combined response" mode.


juleskuehn (approved):
Tested locally and it does the job! Thanks.

PR #103, opened 2024-11-02
ops: StatefulSets, TF security, k8s backups

jannable:
This pull request includes significant updates to the application's infrastructure and deployment process. Key changes include optimizing Docker and Kubernetes configurations, improving database management with statefulsets and backups, enhancing security measures through secret management and OIDC implementation, and streamlining environment setups. The PR also addresses scalability with horizontal pod scaling, adds cron jobs for data synchronization, and makes several adjustments to align with best practices and specific requirements. Additionally, it includes various TODO items for future improvements and removes unnecessary components to streamline the deployment process.

- Changed app path in Docker file so that laws XML could load without further code changes.
- Added db wait, migrate, and collect static to the migrate_collect script.
- Made the static file clean up more concise in the initial_setup script.
- Removed the force option on corporate library command.
- Removed admin from Django settings.
- Moved Django db into the AKS cluster and removed Cosmos DB from the setup.
- Added subnet ranges to app, web, and db. (Although there's nothing in the db subnet anymore. Only the keyvault and storage are in app. Cluster is in web.)
- Added PreProd/Pilot environment settings.
- Adjusted default OpenAI quota limit to match Microsoft defaults.
- Added k for kubectl alias in setup Dockerfile.
- Added TODO items (commented out) for implementing Velero when the change request is approved.
- Updated the link for the content filter form in the README to go to the Canadian form instead of the US form. (Caused delays before.)
- Added a check for the certificate so user can choose to either use a CA-certified cert from a selected keyvault, generate a new one using Let's Encrypt, or skip altogether. (The skip is recommended for updates to avoid reaching the API limit on Let's Encrypt.)
- Moved all the secrets out of Terraform to avoid storing secrets in the Terraform state. They're all done in one script now, and will check to make sure API keys match.
- Removed security context on celery worker. It was causing an issue.
- Added TODO items (commented out) for pgBouncher to address connection pooling against the databases. This will become more important as usage increases. Need to resolve minor obstacle.
- Added database backup cron job. Maintains full daily backups of both databases in Azure storage.
- Added media sync cron job. Maintains backup of user-uploaded files to Azure Storage.
- Converted Django, Redis, and Postgres to statefulset.
- Modified storageclass to enable retention of database data through restarts and node rollovers.
- Add TLS secret to hold CA-signed certificate when that option is selected. Also added script logic to handle fetching it from user-selected keyvault. Note: Further work required to confirm it works.
- Added horizontal pod scaler for postgres databases to enable use of multiple nodes when required.
- Improved k8s run script to check for statefulsets as well as deployments before proceeding.
- Set Terraform storage to encrypted. Uses Microsoft-managed keys. Doesn't hold any secret values anymore.
- Added random 5-digit alphanumeric string to storage account names to ensure uniqueness when running.
- Enabled OIDC issuer and workload identity on AKS so that we can use RBAC more effectively within the cluster. (Will be handy when we use Velero.)
- Resolved issues relating to alerts and added missing tags to avoid Terraform from constantly dropping them.
- Added service endpoints and configured networking to restrict traffic from JUS only.
- Updated content filter properties to the new base policy format.
- Added Data Blob Owner role to the admin group.
- Adjusted the customer managed key config to prevent Terraform from constantly recreating it.
- Added user-identity to the storage to make cmk possible.
- Added much-needed parameters to the scripts
- Added readiness & liveness probes.
- Parameterized the auto-scaling and added horizontal pod scaling.

To run terraform and pass parameters:
`bash run_terraform.sh --env-file .env.dev --subscription OttoDev --skip-confirm y --auto-approve y --enable-debug n`

To run k8s and pass parameters:
`bash run_k8s.sh --env-file .env.dev --subscription OttoDev --cert-choice 3 --init-script n --set-dns-label n`

To run build_and_push and pass parameters:
`./build_and_push_image.ps1 -subscription OttoDev -acr jusdevottoacr -skipNetworkCheck y`


haitmo (dismissed):

django/entrypoint.sh
```diff
@@ -1,7 +1,15 @@
...
+echo "Waiting for database..."
+{ python manage.py wait_for_db || { echo "Error: Wait for database failed"; exit 1; } }
+
+# Run migrations
+echo "Running migrations..."
+{ python manage.py migrate --noinput || { echo "Error: Migrate failed"; exit 1; } }
+
+# Collect static files
+echo "Collecting static files..."
+{ python manage.py collectstatic --noinput --clear || { echo "Error: Collect static files failed"; exit 1; } }
```
juleskuehn, line 13: Migration and collectstatic are already dealt with in "init containers" now, so this can be reverted.

django/entrypoint.sh
```diff
@@ -1,7 +1,15 @@
-#!/bin/sh
-# python manage.py migrate
+#!/bin/bash
 
-# We may not actually want to run then following every time the container is created...
-# bash .devcontainer/post-create.sh
+# Wait for the database to be ready
+echo "Waiting for database..."
+{ python manage.py wait_for_db || { echo "Error: Wait for database failed"; exit 1; } }
```
juleskuehn, line 5: Couldn't we just have a health check on the database service, and have Django pod "depends on" database?

setup/k8s/celery.yaml
```diff
@@ -34,11 +34,6 @@ spec:
       labels:
         pod: celery-worker
     spec:
-      # The Celery worker and beat processes run as a non-root user (UID 1000)
-      securityContext:
-        runAsUser: 1000
```
juleskuehn, line 39: Issues with running Celery as non-root user were successfully resolved?

setup/terraform/modules/aks/main.tf
```diff
@@ -15,15 +15,18 @@ resource "azurerm_kubernetes_cluster" "aks" {
...
   private_cluster_enabled = var.use_private_network
-  private_dns_zone_id     = var.use_private_network ? "System" : null
 
   # Configure the default node pool
   default_node_pool {
-    name       = "default"
-    node_count = 2
-    vm_size    = "Standard_D4s_v3"
+    name           = "default"
+    node_count     = 2
```
juleskuehn, line 27: If we have pods autoscaling, don't we also need this to autoscale? (Or have more nodes)
I realize this may be bound by organizational limits on vCPUs though.

setup/k8s/djangodb.yaml
```diff
@@ -0,0 +1,202 @@
...
+            - name: POSTGRES_MAX_CONNECTIONS
+              value: "500"
+          ports:
+            - containerPort: 5432
+          resources:
+            requests:
+              cpu: "500m"
+              memory: "4Gi"
+            limits:
+              cpu: "500m"
```
juleskuehn, line 49: Should we consider bumping up the CPU and memory limits before scaling horizontally?

setup/k8s/vectordb.yaml
```diff
@@ -61,30 +46,58 @@ spec:
               cpu: "500m"
               memory: "4Gi"
             limits:
-              cpu: "1"
+              cpu: "500m"
```
juleskuehn, line 49: Same comment re: limits. The vector DB in particular can sometimes consume a lot of CPU, when adding a lot of documents to an indexed table (laws refresh). Not critical to address right now though. I expect autoscaling will not work until we have increased vCPU available for the node pool.

setup/terraform/modules/openai/main.tf
```diff
@@ -29,88 +29,104 @@ resource "azurerm_cognitive_account" "openai" {
...
         {
-          name : "hate",
-          allowedContentLevel : "Medium",
-          blocking : false,
-          enabled : false,
-          source : "Prompt"
+          "name" : "Violence",
+          "severityThreshold" : "Low",
+          "blocking" : false,
+          "enabled" : true,
```
juleskuehn, line 55: What is the benefit of having these non-blocking, but enabled content filters? Are we doing anything with the outputs of the content filter? If not I would rather just see the filters disabled.

django/entrypoint.sh
jannable: I think it's more logical to consolidate all the initial/entrypoint scripts in one place. I would rather see it in entrypoint.sh instead. Any objections to removing it from the yaml?

django/entrypoint.sh
jannable: I added it because I encountered an issue during an initial deploy. I think what might have happened is that the postgres service existed and was healthy but wasn't totally ready yet for db connections. The timing caused Django to error out.

django/entrypoint.sh
jannable: Consolidated to a migrate-collectstatic.sh file and kept init-container.

setup/k8s/celery.yaml
jannable: The team agreed to postpone the work of making sure all the processes are running as non-root.

setup/terraform/modules/aks/main.tf
jannable: I've added a prerequisite that DEV's quota limit extend to 16 and other environments go to 64. Then I added an environment variable for this amount and we adjust the max node count based on that and the chosen size. With 64 vCPU and a node size with 4 vCPUs, the max node count would be computed as 16.

setup/k8s/djangodb.yaml
jannable: Add parameters. There is some give but for databases, the limits should not be significantly different than requests.

setup/k8s/vectordb.yaml
jannable: Same response. Also, increased the quota and used ScalingCalculator to find a good balance.

juleskuehn (approved):

PR #105, opened 2024-11-04
fix: more robust cost tracking/stop button behaviour

smcdonald-jus:
Currently, hitting the Stop button in the AI Assistant stops output, but does not display costs. On the backend, by the time the client-side generator object is created, the model deployment has already done everything. Thus, we get the full cost without actually tracking it. This PR changes that in a few ways:

1. *Changes to htmx_stream*. If the user hits Stop mid-stream, it will stop displaying more tokens as before. However, the generator now exhausts in the background, allowing the full cost to be calculated and displayed.
2. *Changes to responses.py*. In any context that requires multiple generators (summarizing or translating multiple documents, per-doc or per-source Q+A), it actually _is_ possible to prevent some model-side generation if you hit the stop button REALLY fast (before all the generators are made). The output will be empty, but after a moment you will see only the costs for what was actually charged (i.e. whichever generators were created before hitting the Stop button). This one may be hard to test without breakpoints and/or great reflexes.


.vscode/launch.json
```diff
@@ -114,7 +114,8 @@
             "args": [
                 "runserver",
             ],
-            "django": true
+            "django": true,
+            "justMyCode": false
```
smcdonald-jus, line 118: This allows you to put breakpoints in source code on VSCode. I've found it useful when trying to figure out how stuff works.

django/tests/chat/test_chat_views.py
```diff
@@ -218,16 +218,18 @@ async def stream_generator():
     )
     # Iterate over the response_stream generator
     final_output = ""
-    first = True
+    message_counter = 0
```
smcdonald-jus, line 221: Previously, hitting the Stop button would cause htmx_stream to terminate AFTER appending one last response. Now, responses only get appended if the stop button hasn't been pressed yet. Thus, I had to change this test to send the Stop message AFTER the "second thing"

django/chat/templates/chat/components/stop_button.html
```diff
@@ -7,6 +7,6 @@
           hx-get="{% url 'chat:stop_response' message_id=message.id %}"
           hx-swap="none">
     <i class="bi bi-stop-circle"></i>
-    <span class="d-none d-md-inline">{% trans "Stop" %}</span>
+    <span class="d-none d-md-inline">{% trans "Stop Output <br> WARNING: stopping may not reduce costs" %}</span>
```
smcdonald-jus, line 10: Previously this said "_will_ not reduce costs", but that's not technically true as there are some circumstances where you can actually reduce costs if you're fast enough (see above)

juleskuehn:
Thanks, this is working well in terms of reporting the cost later, when the response is done!

The stop button should disappear as soon as the response stops, even if it is still processing in the background.

I don't like the large stop button with all the text in it. Alternative:
* keep stop button the same as before
* when response is stopped, append to its text, in user's language and in italics, etc. _Response stopped early. (Costs may still be incurred after stopping.)_

juleskuehn:
Also, please disable the stop button when stopping is impossible (file translation, adding files to Q&A library).

smcdonald-jus:
Regarding disabling the stop button:

The stop button doesn't appear until the file uploads are done, at which point it's possible to prevent some of them for being translated/summarized if you're quick enough. It'll prevent you from getting any of the results since the Stop message will appear instead, but there is the potential to save some costs.

The same applies to Q+A upload, although all you'd be missing there is the "ready for Q+A" message.

django/chat/responses.py
```diff
@@ -159,15 +159,16 @@ def summarize_response(chat, response_message):
...
             if not file.text:
                 file.extract_text(pdf_method="default")
-            responses.append(
-                summarize_long_text(
-                    file.text,
-                    llm,
-                    summary_length,
-                    target_language,
-                    custom_summarize_prompt,
+            if not cache.get(f"stop_response_{response_message.id}", False):
```
smcdonald-jus, line 162: Due to the way the summarize_response logic works, it's still possible to prevent summarization outright by hitting the stop button reasonably quickly after the uploads. In this case, the output will be blank and there will be no cost object created - but I've verified that it doesn't actually go through to the model.

If you're _not_ quick enough, you get the "Response stopped" message and whatever costs accumulated from the summarization jobs that went through, as expected.

django/chat/responses.py
```diff
@@ -525,6 +529,7 @@ async def add_files_to_library():
             responses = [
                 synthesizer.synthesize(query=input, nodes=sources).response_gen
                 for sources in source_groups
+                if not cache.get(f"stop_response_{response_message.id}", False)
```
smcdonald-jus, line 532: Doesn't break anything to include this, but does require ninja skills to reduce costs

django/chat/utils.py
```diff
@@ -224,13 +229,21 @@ def sse_string(message: str, format=True, dots=False, remove_stop=False) -> str:
...
+                    full_message, format=False, dots=dots_html, remove_stop=remove_stop
+                )
                 await asyncio.sleep(1)
                 first_message = False
-            full_message = response
-            if cache.get(f"stop_response_{message_id}", False):
-                break
-            yield sse_string(full_message, format, dots)
+
+            if remove_stop or not cache.get(f"stop_response_{message_id}", False):
```
smcdonald-jus, line 238: Even with remove_stop = True, the stop button will appear VERY briefly before the sse_string logic that removes it goes into effect.
Thus, we have to include it as a condition here in order to bypass the below logic for translation/Q+A uploads.

juleskuehn (approved):

PR #107, opened 2024-11-04
chore: decrease logging

Rifahaziz:
Print statements replaced with loggers in all python files except command classes.


haitmo (approved):
Thanks Rifah!

PR #110, opened 2024-11-05
fix: remove force argument from corpo library load

smcdonald-jus:


jannable:
Can you also remove --force from the initial_setup.sh file

jannable (approved):

PR #114, opened 2024-11-07
fix:  unescaped HTML warning

mcusteau:
BUGS 2921

https://dev.azure.com/BusinessAnalyticsCentre/BAC%20Agile/_workitems/edit/2921

getting the model to write code causes "One of your code blocks includes unescaped HTML" (XSS attack vector)"

### DIFF
--- a/django/chat/utils.py
+++ b/django/chat/utils.py
@@ -86,7 +86,6 @@ def llm_response_to_html(llm_response_str):
         "img": ["src", "alt", "title"],
         "a": ["href", "alt", "title"],
         "pre": ["class"],
-        "code": ["class"],
         "span": ["class"],
     }
     clean_html = bleach.clean(raw_html, allowed_tags, allowed_attributes)


mcusteau:
The warning is caused by the attribute class="language-python" inside the <code> tag. The purpose of this attribute is to create syntax highlighting when displaying python code.  The simplest solution is to remove the class attribute from the <code> tag using bleach. This won't change anything, except now python code won't have syntax highlighting when displayed. Since Otto is a legal chatbot I don't think this is a big deal.

juleskuehn (commented):
Thanks for finding such a simple fix.

Personally, I do appreciate the syntax highlighting - I wonder if there is another way.

Before merging this, could you look into this related ticket:
https://dev.azure.com/BusinessAnalyticsCentre/BAC%20Agile/_boards/board/t/Otto/Projects?System.IterationPath=%40currentIteration&workitem=2871

Using markdown-it in the browser to parse the returned markdown might solve this problem and others at the same time.

PR #117, opened 2024-11-13
feat: improve summarize instructions to reduce hallucination

juleskuehn:
After some tweaks to the default prompts, hallucinations are much lower.
I don't think there is a need to switch default model to "4o" from "4o-mini" (current).

Added "additional instructions" field and "avoid gendered pronouns" checkbox which can help avoid hallucinations.

By default, it will now avoid using gendered pronouns. This hasn't seemed to negatively impact the summaries.

The additional instructions textarea can help with this too (among other things).

e.g., if summary contains incorrect pronouns, you can correct them while still using the default length/language prompts.
image


django/otto/settings.py
```diff
@@ -98,7 +98,10 @@
 OPENAI_EMBEDDING_COST_PER_TOKEN = 0.0004 / 1000
 
 DEFAULT_CHAT_MODEL = "gpt-4o-mini"
-USD_TO_CAD = 1.36
+DEFAULT_QA_MODEL = "gpt-4o-mini"
+DEFAULT_SUMMARIZE_MODEL = "gpt-4o-mini"
+DEFAULT_TRANSLATE_MODEL = "gpt-4o-mini"
+USD_TO_CAD = 1.38
```
smcdonald-jus, line 104: Is there a better way of doing this besides periodically re-hard-coding the exchange rate? Can we maybe link it up to some kind of live tracker?

django/otto/fixtures/cost_types.yaml
```diff
@@ -168,5 +168,5 @@
     description_fr: Appels API au modèle de déploiement global Azure OpenAI GPT-4o-mini (sortie)
     unit_name: tokens
     unit_name_fr: jetons
-    unit_cost: 0.0006
-    unit_quantity: 1000
\ No newline at end of file
+    unit_cost: 0.6
```
smcdonald-jus, line 171: Why has everything been multiplied by 1000?

juleskuehn: This is how the Azure pricing page shows the costs now.

django/chat/templates/chat/components/options_2_summarize.html
```diff
@@ -19,11 +19,21 @@
...
+  <label for="{{ options_form.summarize_instructions.auto_id }}"
+         class="form-label px-3 mt-2 mb-1 small">{% trans "Additional instructions" %}</label>
+  <div class="px-3">{{ options_form.summarize_instructions }}</div>
+
   <label for="{{ options_form.summarize_prompt.auto_id }}"
          class="form-label px-3 mt-2 mb-1 small">
     {% trans "Custom summarize prompt" %}
     <br>
-    {% trans "(overrides length/language)" %}
+    {% trans "(overrides all previous settings)" %}
```
smcdonald-jus, line 36: Is this still necessary, if it overrides "additional instructions"? What would be an example of behaviour that wouldn't be well-encoded in "Additional Instructions", but would be better enforced with an override prompt here?

django/chat/utils.py
```diff
@@ -327,53 +327,111 @@ def summarize_long_text(
...
+<document>
+{docs}
+</document>
+<instruction>
+"""
+            + f"{custom_prompt}\n</instruction>"
         )
     else:
         length_prompt_template = length_prompts[length][target_language]
+        if gender_neutral:
```
smcdonald-jus, line 425: An interesting thing I noticed from playing around with elephants.msg and some small .txt files I made: including the gender-neutral instructions tends to dilute the "restate the document verbatim" instruction for short docs. You'll see "elephants are wonderful" when the gender-neutral instructions aren't included, and "the document celebrates elephants for being wonderful" when they are. Ditto for single-sentence .txt docs below a certain length.

django/chat/templates/chat/components/options_2_summarize.html
smcdonald-jus: After taking a closer look through the summarize_long_text code, I better understand the function of the custom/override prompt, but I still wonder if it's presented in a way where the benefit is obvious to users. Should we include an example of a summarization prompt as a placeholder or something like that?

django/chat/utils.py
juleskuehn: Hmm, I worried this would be the case. It might be better to just include the gender neutral instruction all the time, and PRIOR to the "repeat verbatim if short" instruction.
Unfortunately, the more instructions you add, the less it cares about any of them.

django/chat/templates/chat/components/options_2_summarize.html
juleskuehn: Let's improve the UX later by having a toggle between "basic/advanced" or something.

smcdonald-jus (approved):

PR #119, opened 2024-11-14
fix: chatfile upload bug; feat: show laws refresh date; misc,

juleskuehn:
See individual commits which are scoped to the small tickets in DevOps


django/text_extractor/templates/text_extractor/completed_documents.html
```diff
@@ -82,10 +82,13 @@
...
 {% if user_request_id %}
-  {# Push URL to history for this specific user request #}
-  <script>history.replaceState(null, '', "{% url 'text_extractor:poll_tasks' user_request_id=user_request_id %}");</script>
-{% endif %}
-{% if refresh_on_load %}
-  {# Refresh the page immediately #}
-  <script>window.location.reload();</script>
+  <script>window.location.href = "{% url 'text_extractor:poll_tasks' user_request_id=user_request_id %}";</script>
 {% endif %}
+<script>
```
smcdonald-jus, line 86: Currently, submitting a new document after processing a previous one takes you to a new page, with your old docs not listed. You can return to them with the Back button (or by using the hash URL if you somehow remembered it), but maybe it's worth having more clear navigation options for users to return to prevoiusly-processed files?

django/librarian/utils/process_engine.py
```diff
@@ -270,9 +270,11 @@ def pdf_to_text_pdfium(content):
     pdf = pdfium.PdfDocument(content)
     for i, page in enumerate(pdf):
         text_page = page.get_textpage()
-        text += f"<page_{i+1}>\n"
-        text += text_page.get_text_range() + "\n"
-        text += f"</page_{i+1}>\n"
+        text_content = text_page.get_text_range()
```
smcdonald-jus, line 273: Could this affect pdf's where pages contain only OCR content? Like, if we completely omit pages without text but there's other content on there that can be OCR'd, will we still be able to get it?

django/chat/static/chat/js/scripts.js
```diff
@@ -291,7 +291,7 @@ class FileUpload {
   }
 
   initFileUpload(i) {
-    var file = this.input.files[i];
+    const file = this.input.files[i];
```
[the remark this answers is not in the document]
juleskuehn: It slows down the page. We should use it only when required by a dependency like DataTables.

django/chat/forms.py
```diff
@@ -27,8 +27,8 @@
 ]
 TEMPERATURES = [
     (0.1, _("Precise")),
-    (0.7, _("Balanced")),
-    (1.2, _("Creative")),
+    (0.5, _("Balanced")),
+    (1.0, _("Creative")),
```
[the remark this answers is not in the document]
juleskuehn: It used to be 0...2. They changed it.

django/text_extractor/templates/text_extractor/completed_documents.html
juleskuehn: Yeah, this was a compromise required for the Celery task polling. My hope is that the back navigation is clear enough.

django/librarian/utils/process_engine.py
juleskuehn: This is actually why I did this change. If there is no content in the whole document detected by this (non-OCR) method, it will fall back to using Azure OCR. Whereas if we added the empty `<page>` tags, the check on content length (to trigger forcing OCR) didn't work properly.

django/chat/templates/chat/components/options_2_summarize.html
```diff
@@ -22,7 +22,7 @@
   <div class="mx-3 mt-2 form-switch form-check">
     {{ options_form.summarize_gender_neutral }}
     <label class="form-check-label small ms-1"
-           for="{{ options_form.summarize_gender_neutral.auto_id }}">{% trans "Avoid gendered pronouns" %}</label>
+           for="{{ options_form.summarize_gender_neutral.auto_id }}">{% trans "Avoid personal pronouns" %}</label>
```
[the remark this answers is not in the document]
juleskuehn: If you look at the prompt (and how it behaves), it discourages using pronouns at all.

smcdonald-jus (approved):

PR #121, opened 2024-11-15
chore: otto branding text; misc small fixes

juleskuehn:
Proposed (this PR):
image

Formerly:
image

A stakeholder wanted a more generic statement that didn't emphasize "legal processes".


django/otto/models.py
```diff
@@ -86,6 +86,10 @@ def roles(self):
     def display_total_cost(self):
```
smcdonald-jus, line 86: Looks like we no longer need this function anymore, since it's not used anywhere?

smcdonald-jus (approved):

PR #122, opened 2024-11-15
feat: automatic updating of USD-to-CAD exchange rate

smcdonald-jus:
Rather than haivng a hardcoded value for the currency conversion (which would have to be manually updated regularly), we can pull the value from the Bank of Canada website at regular intervals. Using a monthly average here.

Note: there are dedicated exchange rate API's, but they're either (a) not free or (b) blocked on the JUS network.


django/tests/otto/test_exchange_rate_update.py
```diff
@@ -0,0 +1,12 @@
...
+from otto.tasks import update_exchange_rate
+
+@pytest.mark.django_db()
+def test_exchange_rate_update(client, all_apps_user):
+    update_exchange_rate()
+    otto_status = OttoStatus.objects.singleton()
+    saved_exchange_rate = otto_status.exchange_rate
+    decimal_places = str(saved_exchange_rate).split('.')[-1] 
+    assert isinstance(saved_exchange_rate, float)
+    assert len(decimal_places) < 5 and len(decimal_places) > 1
```
juleskuehn, line 12: Are you certain that a rate such as 1.4 (1 decimal place only) or 1.382948 won't occur?

juleskuehn (approved):
Tested script with `python manage.py update_exchange_rate` and tested conversion display in browser.

Works great! Feel free to merge if my comment is a non-issue.

django/tests/otto/test_exchange_rate_update.py
juleskuehn: Perhaps replace this with an assert that it is within the range of 0 and 2?

smcdonald-jus: The Bank of Canada seems to give a value with 4 decimal places, and the old hardcoded value was 2 decimal places. I guess I'm making an assumption here that the BoC will always display 4 decimal places, but it does so for time periods ranging from a week to a year so it seems reasonable.

PR #124, opened 2024-11-18
fix: error handling/reporting (chat, librarian)

juleskuehn:
1. Add try/except around otto_response to ensure a `StreamingHttpResponse` is always returned with an error message and an "Error ID" (random 7 characters) for checking against logs:
image
image

2. Same in Librarian:
image
image

3. Fix: Queue'd files in librarian status updating (previously, the status "QUEUED" was excluded from polling)
4. Fix: QA "read full documents" mode wasn't working with "Entire library" selected before.

---

When settings.DEBUG is True, you see more verbose errors right in Otto UI:
image


haitmo (approved):
Thanks for making the changes!

smcdonald-jus (approved):

PR #128, opened 2024-11-21
feat: Reintroduce and refine response evaluation

smcdonald-jus:
Reintroducing the old eval_responses methodology to allow for evaluation of LLM performance as we tweak things. Some refactoring was required to get it working again.

Also wrote a small number of test examples now that we don't have access to the same docs as before.


django/chat/management/commands/eval_responses.py
```diff
@@ -255,9 +255,9 @@ def _create_test_user():
 def _test_qa_response(eval_instance, user):
     try:
         # Create chat and build chat history
-        chat = Chat(user=user)
+        chat = Chat.objects.create(user = user, mode = eval_instance["mode"])
```
smcdonald-jus, line 258: The original version had `chat = Chat(user=user)`. That bypassed the custom create method in ChatManager, so the resulting object had no security label which broke things upstream.

django/chat/management/commands/eval_responses.py
```diff
@@ -255,9 +255,9 @@ def _create_test_user():
...
         chat.save()
-        chat_options = ChatOptions.objects.create(chat=chat, mode=eval_instance["mode"])
+        chat_options = ChatOptions.objects.get(chat=chat)
```
smcdonald-jus, line 260: With the above change, the custom create method creates the ChatOptions object by default, so you only have to get it here.

django/chat/eval/glyph_qa.yaml
```diff
@@ -0,0 +1,36 @@
+
+- mode: qa
```
smcdonald-jus, line 2: Wrote the responses myself based on some preliminary 4o-mini output, and my own understanding of the source material. Obviously you don't want to use pure LLM output as a metric for LLM performance.
Would like a second set of eyes here in particular since I'm no typography expert.

django/chat/eval/glyph_qa.yaml
```diff
@@ -0,0 +1,36 @@
...
+  vector_store_table: Corporate
+  history:
+  - user: When might a character be comprised of multiple glyphs?
+  responses: 
+  - A character is made of multiple glyphs in situations such as the use of diacritics to modify a base character (e.g., ⟨à⟩ consisting of the letter "a" and a grave accent), or when using older typerwriters (e.g. needing to combine an apostrophe and period to create an exclamation mark).
+  sources:
+  - In most languages written in any variety of the [Latin alphabet](/wiki/Latin_alphabet "Latin alphabet") except English, the use of [diacritics](/wiki/Diacritic "Diacritic") to signify a sound mutation is common. For example, the grapheme ⟨à⟩ requires two glyphs
+  - Older models of [typewriters](/wiki/Typewriter "Typewriter") required the use of multiple glyphs to depict a single character, as an overstruck [apostrophe](/wiki/Apostrophe "Apostrophe") and [period](/wiki/Full_stop "Full stop") to create an [exclamation mark](/wiki/Exclamation_mark "Exclamation mark").
+
+- mode: qa
```
smcdonald-jus, line 31: These tests are pretty limited in number, and likely not very robust. At some point we should put more effort into thinking harder about how to "break things", more thoroughly test the stuff that matters, etc.

django/requirements.txt
```diff
@@ -75,6 +75,7 @@ pypdf==5.1.0
 reportlab==4.2.0
 pyreadstat==1.2.7
 openpyxl==3.1.5
+matplotlib==3.9.2
```
smcdonald-jus, line 78: Necessary for the plots in the notebook

django/chat/eval/glyph_qa.yaml
juleskuehn: Agreed, I care less about the content of the tests right now (in this PR) than the ability to run them.

django/chat/management/commands/eval_responses.py
juleskuehn: Great, thanks. There shouldn't be spaces around the keyword arguments = though. Can you make sure your format-on-save is working?

django/chat/management/commands/eval_responses.py
```diff
@@ -269,7 +269,7 @@ def _test_qa_response(eval_instance, user):
                 )
             if "vector_store_table" in eval_instance:
                 library = Library.objects.get(
-                    vector_store_table=eval_instance["vector_store_table"]
+                    name=eval_instance["vector_store_table"]
```
juleskuehn, line 272: For clarity, the string "vector_store_table" should be replaced with "library_name" in all these files.

django/requirements.txt
juleskuehn: Put this in requirements_dev.txt instead so it doesn't increase the prod container size.

juleskuehn (commented):
Looks good Shaun! A few nitpicks, see below.

django/chat/management/commands/eval_responses.py
smcdonald-jus: Like, in the library model too? Just replace the string across the whole codebase?

juleskuehn (approved):

PR #130, opened 2024-11-22
feat: QA on combined full documents

juleskuehn:
Combined full documents Q&A mode allows for comparison queries (etc) that require multiple, full documents in the LLM context.

image

This also fixes a few other bugs in QA mode, and increases Celery concurrency.


django/chat/responses.py
```diff
@@ -420,25 +422,39 @@ async def add_files_to_library():
...
             filter_documents = Document.objects.filter(
                 data_source__library=chat.options.qa_library
             )
         document_titles = [document.name for document in filter_documents]
-        summary_responses = [
-            llm.tree_summarize(
-                context=document.extracted_text,
+        if chat.options.qa_mode == "summarize_combined":
+            # Combine all documents into one text, including the titles
+            combined_documents = "\n\n---\n\n".join(
```
smcdonald-jus, line 433: Does this *always* guarantee that the AI will properly delineate documents? I'm imagining a situation where one of the docs contains this same pattern of line breaks and bold text, which would presumably be falsely interpreted as another document separation.

django/locale/translation/translations.json
```diff
@@ -3982,5 +3982,13 @@
     "An error occurred.": {
         "fr": "",
         "fr_auto": "Une erreur s’est produite."
+    },
+    "Read each document separately  (slow, $$)": {
+        "fr": "",
+        "fr_auto": "Lisez chaque document séparément (lent, $$)"
+    },
+    "Read all documents at once (slowest, $$)": {
```
smcdonald-jus, line 3990: I suggest slightly different wording here. Because the per-document responses stream "in parallel", they appear to all get read/answered "at once". Maybe the following wording is more clear?

"Answer for each document separately (slow, $$)"
"Compare all documents in single answer (slowest, $$)"

django/chat/views.py
```diff
@@ -642,14 +642,6 @@ def _copy_options(source_options, target_options):
         chat_options = chat.options
         post_data = request.POST.copy()
 
-        # In case of duplicate values, remove them by taking the first value from each list
```
smcdonald-jus, line 645: Why remove this, out of curiosity?

django/chat/responses.py
juleskuehn: Fair. I should probably change this to XML style.

django/locale/translation/translations.json
juleskuehn: Good point. I wanted to make it clear that the documents are read in their entirety. I'll try to think of a wording that captures both (while remaining succinct).

django/chat/views.py
juleskuehn: I honestly don't know why this was in here to begin with.
It breaks the saving of anything with multiple values! (specifically, the "data sources", "documents" multiselects).

Rifahaziz (approved):

PR #132, opened 2024-11-22
feat: chat title in <title>; title case apps; js fixes

juleskuehn:


django/chat/static/chat/js/scripts.js
```diff
@@ -180,7 +180,7 @@ document.addEventListener("DOMContentLoaded", function () {
 });
 // On prompt form submit...
 document.addEventListener("htmx:afterSwap", function (event) {
-  if (event.target.id != "messages-container") return;
+  if (event.detail?.target?.id != "messages-container") return;
```
juleskuehn, line 183: Being more selective removes redundant execution of the code that follows.

django/chat/static/chat/js/scripts.js
```diff
@@ -519,3 +519,13 @@ document.addEventListener("DOMContentLoaded", function () {
     }
   });
 });
+
+
+function updatePageTitle(title = null) {
```
juleskuehn, line 524: Handles first page load, title update after streaming response, and manual renaming

django/chat/templates/chat/components/chat_list_item_title_edit.html
```diff
@@ -22,12 +22,15 @@
 
 {% block page_script %}
     <script>
-        htmx.onLoad((element) => {            
-            if(element instanceof HTMLFormElement) {                
+        htmx.onLoad((element) => {
+            if(element instanceof HTMLFormElement) {
                 const input = document.getElementById('{{ form.title.id_for_label }}');                
-                input.setSelectionRange(-1, -1);
```
juleskuehn, line 28: This was broken, and is addressed better with onfocus in the input element (see forms.py)

PR #135, opened 2024-11-25
feat: lex experiment prototype

Rifahaziz:
Michel and I have worked on getting a new application prototyped that uses the text extractor and Q&A functionalities. The application helps LEX users extract important information from scanned documents of Notice of Appeals.


.devcontainer/Dockerfile_celery
```diff
@@ -5,6 +5,9 @@ RUN apt-get update && apt-get install -y \
     && apt-get clean \
     && rm -rf /var/lib/apt/lists/*
     
+# Verify the installation of pdfinfo
+RUN pdfinfo -v
+
```
Rifahaziz, line 10: verifying the pdfinfo is installed with poppler, because sometimes it just isn't.

django/lex_experiment/models.py
```diff
@@ -0,0 +1,34 @@
...
+    celery_task_ids = models.JSONField(default=list, blank=True, null=True)
+    access_controls = models.ManyToManyField(
+        "otto.AccessControl",
+        related_name="lex_experiment_outputfile_controls",  # Unique related_name
+        blank=True,
+    )
+    answers = models.JSONField(default=list, blank=True, null=True)
+
+    def get_permission_parents(self):
+        return [self.user_request]
```
Rifahaziz, line 34: we might need the files later on depending what the users want so I've kept the pdf and txt files here.

.devcontainer/Dockerfile_celery
juleskuehn: What should happen in the case that it isn't?

Rifahaziz: It can't find pdfinfo method that's required to convert pdf into image pages, its called with poppler. So it will run into an error saying it can't find pdfinfo().

juleskuehn (commented):
I tested it out with some files Michel sent me, and I think this is in good shape for the Lex team to try out!

A couple suggestions for user experience:
1. "Upload scanned PDFs or images" heading:
  How about "Upload Notice(s) of Appeal"
  And rename the button from "Submit" to "Extract information"
2. With a bit of HTML/CSS changes, you can get the table much wider which might be nice in this case.
3. I don't like the "Status/Cost" on the separate row - the table headers don't correspond to it, and if they copy/pasted the table it would have these extra rows. If there is more room after the table width change, can you put that in a "Status" column?

Other requests:
1. Needs `makemigrations` run to add the costing "feature".
2. Text casing: change to "Information extracted" (lowercase e), "LEX Experiment" (uppercase "LEX", uppercase "E") throughout, e.g. in apps.yaml. Update translations after.
3. This seems to duplicate a lot of the code from Text Extractor. When the code is identical, can you import the models and/or functions from `text_extractor.models`, `text_extractor.tasks`, `text_extractor.utils` etc? This would also improve your test coverage since much of that code already has tests. (Rifah, I think you were planning to add more tests to the text extractor code too?)

django/lex_experiment/templates/lex_experiment/lex_experiment.html
```diff
@@ -24,7 +24,7 @@
           hx-indicator="#spinner"
           onsubmit="document.getElementById('submit').disabled = true;">
       {% csrf_token %}
-      <h2 class="h3" id="upload-label">{% trans "Upload scanned Notice(s) of Appeal" %}</h2>
+      <h2 class="h3" id="upload-label">{% trans "Upload Scanned Notice(s) of Appeal" %}</h2>
```
juleskuehn, line 27: They don't have to be scanned for this to work - If there's a chance this might confuse the users, just remove the word "scanned".

juleskuehn (dismissed):
Changes look good Rifah! I made one comment. If you make the change I can quickly re-approve.

Rifahaziz:
All done now. thank you for the feedback Jules!

juleskuehn (approved):
Looks good!

PR #138, opened 2024-11-26
fix: preset permissions

mcusteau:
https://dev.azure.com/BusinessAnalyticsCentre/BAC%20Agile/_workitems/edit/3005

- removed the ability of admins to see private presets of users (as requested by Jules)
- added rules regarding presets in rules.py
- added more unit tests for presets


haitmo (changes requested):
Thanks Michel.

One change is required. The modified button should not appear when a user is not the owner. You can reuse the rule you created.

image

haitmo (approved):

PR #139, opened 2024-11-26
ops: Add service endpoints to subnets

jannable:


juleskuehn:
@jannable Does this belong in the otto-infrastructure repo?

PR #140, opened 2024-11-27
refactor: Q&A mode switching changes scope

mcusteau:
https://dev.azure.com/BusinessAnalyticsCentre/BAC%20Agile/_workitems/edit/3012

- changing search mode in Q&A should not change the scope
- Exception, if we have Entire Library scope already selected and then we choose Full Documents search mode, switch scope to Select Documents so the users don't accidentally rack up costs.


mcusteau:
the function limitScopeSelect is what is causing the scope to change. I removed it from the call when the search mode changes. We still need the function to get called when we change library.

PR #143, opened 2024-11-27
ops: Private networking with jumpbox VM

jannable:


juleskuehn:
@jannable Does this PR need to move to the otto-infrastructure repo now?

PR #151, opened 2024-12-02
feat: otto logo

juleskuehn:


django/otto/templates/base.html
```diff
@@ -43,11 +43,14 @@
...
-                   height="30"
+              <img src="{% static 'otto-icon.png' %}"
+                   width="27"
+                   height="auto"
                    class="d-inline-block align-top"
-                   alt="Otto logo" />
+                   style="position: relative;
+                          top: 4px;
+                          margin-right: 3px"
+                   alt="{% trans 'Smiling robot head' %}">
```
Rifahaziz, line 53: when I click on the logo, it moves down and touches the "t" of otto. Refreshing moves it up again to the origin place. Is this movement supposed to happen? Just the height position is changing when you hover over the full-body logo.

juleskuehn: Yeah, it is intentional, so that you can reach the search bar with the mouse. If it's too weird I can get rid of the animation and just have the logo be lower from the get-go. I was trying to implement Jason's suggestion.

Rifahaziz (approved):

PR #153, opened 2024-12-03
chore: add further tests 

Rifahaziz:
Added tests for text extractor, celery polling, laws_utils and chat_translate. Coverage is 88% without template wizard now.


django/tests/conftest.py
```diff
@@ -228,3 +265,23 @@ def new_feedback_form(user):
         return feedback
 
     return new_feedback_form
+
+
+@pytest.fixture
```
smcdonald-jus, line 270: Does this fixture actually get used anywhere? It looks like the tests that contain something called output_file define it independently with MagicMock(). Am I missing something?

Rifahaziz: yes, this output_file fixture is used in test_download_document method. Basically tried to create a simple output_file without requiring accesskey and userrequest here. But for test_poll_tasks, we need the actual output file as things get updated on the real outputfile object.

smcdonald-jus: Oop, I missed that. My bad!

smcdonald-jus (approved):

PR #154, opened 2024-12-05
chore: update requirements to resolve openai bug

juleskuehn:
Also updates other dependencies as suggested by dependabot.


amusbag (approved):
All tests passed locally as well.

PR #155, opened 2024-12-06
refactor: save button to save and update presets

mcusteau:
https://dev.azure.com/BusinessAnalyticsCentre/BAC%20Agile/_workitems/edit/2954

- When users have a preset loaded in the chat and they click they save button, they are asked if they want to update the current preset or save a new one

- Loading a preset is currently broken, this PR also has a fix for the issue

- After a user creates a new preset, instead of being redirected to the preset list where they now have to click on the new preset to load it, the new preset upon creation is simply automatically loaded.


django/chat/models.py
```diff
@@ -274,6 +277,29 @@ def get_accessible_presets(self, user: User, language: str = None):
             .order_by(*ordering)
         )
 
+    def get_created_presets(self, user: User, language: str = None):
```
haitmo, line 280: Is this function still necessary with your new way of doing things?

django/chat/static/chat/style.css
```diff
@@ -1243,6 +1243,21 @@ ul#preset-edit-tabs {
   }
 }
 
+.preset-form-user-choice .card-button:hover {
```
haitmo, line 1246: I think these can be removed as well no?

haitmo (commented):
Thanks Michel!

When hovering over the options the cards are being resize slightly, would you be able to fix that?

image

django/chat/views.py
```diff
@@ -619,14 +635,52 @@ def chat_options(request, chat_id, action=None, preset_id=None):
...
+                        "prompt": preset.options.prompt,
+                        "preset_loaded": "true",
+                        "preset_id": preset.id,
+                    },
+                )
 
+        return HttpResponse(status=500)
+    elif action == "update_preset":
+        preset = get_object_or_404(Preset, id=preset_id)
+        _copy_options(chat.options, preset.options)
```
haitmo, line 662: I don't think this exists any more. I believe it was moved to chat utils.py

haitmo: Could you add more tests for the new code you added?

django/cmctl-tmp/cmctl.tar.gz
haitmo, line 1: what's this?

haitmo (commented):
Could you also run load_app_translation?

haitmo (approved):

PR #158, opened 2024-12-09
feat: cost dashboard date filters & CSV download

juleskuehn:


django/chat/llm.py
```diff
@@ -237,7 +237,7 @@ def _get_llm(self) -> AzureOpenAI:
 
     def _get_embed_model(self) -> AzureOpenAIEmbedding | MockEmbedding:
         if self.mock_embedding:
-            return MockEmbedding(1536, callback_manager=self._callback_manager)
+            return MockEmbedding(1536)
```
smcdonald-jus, line 268: Guessing the callback_manager argument was found to be redundant or something?

django/otto/views.py
```diff
@@ -670,9 +679,83 @@ def cost_dashboard(request):
...
+    start_date = request.GET.get("start_date", None)
+    end_date = request.GET.get("end_date", None)
 
-    # First, filter the costs by the selected options
     raw_costs = Cost.objects.all()
+
+    # Filter by dates
+    if date_group == "last_90_days":
+        start_date = timezone.now().date() - timedelta(days=90)
+        end_date = timezone.now().date() - timedelta(days=1)
```
smcdonald-jus, line 691: What's the rationale behind excluding the current date from the "last N days" date groups?

juleskuehn: The current day is incomplete. So it gives you a slightly false idea of what "a week" costs.

smcdonald-jus: Fair enough!

smcdonald-jus (approved):

PR #162, opened 2024-12-11
fix: auto-stop / start chat scrolling when user scrolls

juleskuehn:
Also fixes a small bug with response stopping.


amusbag (approved):
Tested locally and works!

PR #165, opened 2024-12-11
fix: reduce number of near-simultaneous LLM calls

smcdonald-jus:
Limits the number of "simultaneous" LLM calls to 5 by creating batches for per-doc/per-source Q+A, and document summaries.

Also added some small fixes to tests, and allowed for the streaming dots to appear when waiting for batches.


django/tests/conftest.py
```diff
@@ -144,6 +144,7 @@ def mock_pdf_file2():
 
     with open(filename, "rb") as f:
         yield f
+    os.remove(filename)
```
smcdonald-jus, line 147: temp_file2.pdf wasn't removed after running tests locally, so you always had to manually delete it before commits

django/tests/otto/test_cleanup.py
```diff
@@ -249,8 +249,6 @@ def test_delete_old_chats_task(client, all_apps_user):
     chat = user.chat_set.first()
     chat_id = chat.id
     assert chat is not None
-    # Check that the chat.accessed_at is close to the start time
-    assert (chat.accessed_at - start_time).total_seconds() < 2
     time.sleep(3)
```
smcdonald-jus, line 252: Removed the part of this test that assumed it'd take less than 2 seconds to access a chat. It _usually_ would, but I figure it's neither consistent enough nor informative enough to be worth keeping.

django/chat/responses.py
```diff
@@ -178,12 +180,26 @@ def summarize_response(chat, response_message):
                         instructions,
                     )
                 )
+
+        title_batches = batch(titles, 5)
```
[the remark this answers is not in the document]
juleskuehn: Would you mind making this a named variable `batch_size`, and just rename the function "batch" to a verb phrase like "create_batches" so it is extra clear, `title_batches = create_batches(titles, batch_size)`

django/chat/utils.py
```diff
@@ -477,6 +477,12 @@ def get_source_titles(sources):
     ]
 
 
+def batch(iterable, n=1):
+    l = len(iterable)
+    for ndx in range(0, l, n):
+        yield iterable[ndx : min(ndx + n, l)]
```
juleskuehn, line 483: Please don't use lowercase `l` as a variable name; I had to read this 10 times

django/chat/utils.py
```diff
@@ -529,6 +535,35 @@ async def combine_response_replacers(generators, titles):
...
+            # relevant was found in the batch, just retain previous batches
+            yield stream
+            await asyncio.sleep(0)
+        else:
+            final_streams.append(response)
+
+    if not final_streams and pruning:
+        # If we're pruning (combine_response_generators only) and have nothing after
+        # iterating through all batches, stream the pruning message again
+        yield "**No relevant sources found.**"
```
juleskuehn, line 564: Needs translation

django/chat/utils.py
```diff
@@ -529,6 +535,35 @@ async def combine_response_replacers(generators, titles):
...
+        async for response in generator:
+            if stream:
+                # Add line between already-streamed batches and streaming batch
+                stream_value = stream + "\n\n---\n\n" + response
+            else:
+                # Don't need line if there's nothing streamed
+                stream_value = response
+            yield stream_value
+            await asyncio.sleep(0)
+        if pruning and response == "**No relevant sources found.**":
```
juleskuehn, line 553: Needs translation

django/tests/conftest.py
juleskuehn: Thanks!

django/chat/utils.py
smcdonald-jus: Hahaha, my bad. This was copied from StackExchange

juleskuehn (approved):
Everything working as it should now. Thanks Shaun. Good to merge.

PR #168, opened 2024-12-12
test: include celery tests in GH actions; exclude template_wizard from coverage

juleskuehn:


.github/workflows/pytest.yml
```diff
@@ -43,6 +43,10 @@ jobs:
         ports:
           # Maps tcp port 5432 on service container to the host
           - 5432:5432
+      redis:
+        image: redis:7.0.15-alpine3.20
```
haitmo, line 47: Minor. Should we match our redis image or at least version (currently redis:7.0.11-bullseye) with what's in k8s?

haitmo (approved):

.github/workflows/pytest.yml
juleskuehn: Fair point. This is the same image we use in local dev. The reason for alpine image in local and github is the smaller size, vs. bullseye for better performance in k8s.
I hope we can eventually also run tests in the kubernetes dev cluster.

PR #169, opened 2024-12-12
fix: load preset where library isn't accessible

juleskuehn:
Previously, loading a ChatOptions preset that referenced either:
* a non-existent library (i.e. one that had been deleted since the preset was created)
* an inaccessible library (i.e. one that the user doesn't have access to)
would silently fail.

Same with when using "reset" button in the ChatOptions sidebar.

There actually was already a check to fix this for the "chat" view (initial loading of a chat) but it was only partial.

Tests added to illustrate issue and confirm functionality.


django/chat/models.py
```diff
@@ -101,7 +101,7 @@ def from_defaults(self, mode=None, chat=None):
             from chat.utils import copy_options
 
             new_options = self.create()
-            copy_options(chat.user.default_preset.options, new_options)
+            copy_options(chat.user.default_preset.options, new_options, chat.user)
```
juleskuehn, line 104: When a new chat is created, the user might have a default preset selected which has a QA library that isn't accessible to them. We do this permission check inside `copy_options`.

django/chat/views.py
```diff
@@ -209,14 +209,13 @@ def chat(request, chat_id):
...
-        form = ChatOptionsForm(instance=chat.loaded_preset.options, user=request.user)
-    else:
-        form = ChatOptionsForm(instance=chat.options, user=request.user)
+    if not chat.options.qa_library or not request.user.has_perm(
+        "librarian.view_library", chat.options.qa_library
+    ):
+        # The copy_options function fixes these issues
+        copy_options(chat.options, chat.options)
+
+    form = ChatOptionsForm(instance=chat.options, user=request.user)
```
juleskuehn, line 218: As discussed with Michel, this is actually the preferred behaviour.

django/chat/views.py
```diff
@@ -542,7 +541,7 @@ def chat_options(request, chat_id, action=None, preset_id=None):
         # Update the chat options with the preset options
         copy_options(preset.options, chat.options)
 
-        chat_options_form = ChatOptionsForm(instance=preset.options, user=request.user)
+        chat_options_form = ChatOptionsForm(instance=chat.options, user=request.user)
```
juleskuehn, line 544: Because copy_options may have reset the QA library, its now important to pass the updated chat.options instead of the source preset.options

django/librarian/templates/librarian/components/edit_details.html
```diff
@@ -1,7 +1,7 @@
 {% load i18n %}
 
 <div id="librarian-details" {% if swap %}hx-swap-oob="true"{% endif %}>
-  {% include "components/messages.html" %}
+  {# {% include "components/messages.html" %} #}
```
juleskuehn, line 4: Commented this out because the messages also show up as toasts in the background.

TODO: Get the toasts to have a higher z-index (past the lightbox)

django/otto/templates/base.html
```diff
@@ -100,7 +100,6 @@
     {% endif %}
 
     <div aria-live="polite" aria-atomic="true" class="sticky-top">{% include "components/toasts.html" %}</div>
-    {% include "components/messages.html" with add_padding="true" %}
```
juleskuehn, line 103: Don't need this now that we have the toasts in base.html

mcusteau (approved):
Works well, if possible make the message that appears for the missing library last a little bit longer so the user has time to read it

juleskuehn:
> Works well, if possible make the message that appears for the missing library last a little bit longer so the user has time to read it

Added a ticket for this.

PR #170, opened 2024-12-13
feat: url allow-list

juleskuehn:
image

image

Worth noting, the message is not *entirely* true, but without referencing "domains and subdomains", it isn't easy to clarify that "justice.gc.ca" is allowed but "abcdefgc.ca" is not allowed.


haitmo (approved):

django/chat/utils.py
```diff
@@ -574,6 +575,7 @@ def sort_by_max_score(groups):
 
 
 def change_mode_to_chat_qa(chat):
+    chat.options.mode = "qa"
```
juleskuehn, line 578: Fixes a minor pre-existing bug.

django/librarian/templates/librarian/components/edit_details.html
```diff
@@ -1,7 +1,7 @@
 {% load i18n %}
 
 <div id="librarian-details" {% if swap %}hx-swap-oob="true"{% endif %}>
-  {# {% include "components/messages.html" %} #}
+  {% include "components/messages.html" %}
```
juleskuehn, line 4: Until the toasts appear above the lightbox overlay, it makes sense to continue showing the messages inside the librarian modal.

django/librarian/views.py
```diff
@@ -86,7 +86,10 @@ def modal_view(request, item_type=None, item_id=None, parent_id=None):
                 show_document_status = True
             else:
                 logger.error("Error updating document:", errors=form.errors)
-                selected_data_source = get_object_or_404(DataSource, id=parent_id)
+                selected_data_source = (
+                    DataSource.objects.filter(id=parent_id).first()
+                    or form.instance.data_source
+                )
```
juleskuehn, line 92: Found this bug as I was testing editing document URLs in the librarian modal.

haitmo (approved):

PR #172, opened 2024-12-16
ops: added localization on ci.

haitmo:
Runs localization command in prs.

Will create a commit (as GitHub bot) if translations.json is updated.

image


juleskuehn (approved):
Looks good to me. If it causes issues we can revert or consider changing the trigger to "push to main".

juleskuehn (approved):

PR #173, opened 2024-12-17
feat: public example presets

mcusteau:


juleskuehn:
The logic should also support more example presets in the YAML (any number) but this is a good start. I had to make some changes to protect against errors - particularly modifying or deleting the global preset. Still need more checks on the back-end.

The bilingual aspects are causing issues with user-created presets. i.e. saving when in French UI will delete the English text - switching back to English it is now blank.

I suggest that we instead provide a separate, French default preset ("Otto default (French)") - which can easily be selected as the user's default - then go back to unilingual ChatOptions etc. The Otto default French can still be selected programmatically in ChatOptions.from_default() method. The only situation this doesn't address is when a user switches language and DOESN'T create a new chat. I don't intend to address this edge case. It makes more sense that the presets are unilingual, because we have a lot of bilingual users who may want to work on tasks in both languages, while keeping the user interface in their preferred language.

I will work on these changes over the next few days.

juleskuehn:
Although tests pass there is in fact an issue with loading presets. The current chat options is not correctly updated in the database. However, the preset *appears* to load in the sidebar.

amusbag (approved):
Fixed resetting of QA autocomplete on edit libraries modal close.

All other changes worked fine.

Approved!

PR #175, opened 2024-12-18
fix: cost dashboard chart disappearing

amusbag:
1. fix: an issue where the chart in the cost dashboard would disappear on filter change and the page would have to be refreshed in order for the chart to reappear.
2. feat: shifted the days to include the current day for "Last 7 days", "Last 30 days" and "Last 90 days" filter.
3. updated privacy notice as requested.

### DIFF
--- a/django/chat/templates/chat/components/chat_welcome.html
+++ b/django/chat/templates/chat/components/chat_welcome.html
@@ -28,9 +28,12 @@
               <strong>{% trans "No sensitive information" %}:</strong>
               {% trans "Do not upload or process sensitive, confidential, or legally protected documents. This includes personally identifiable information, protected B information, and information protected under solicitor client privilege." %}
             </p>
-            <p class="mb-0">
+            <p>
               <strong>{% trans "Not for legal advice or making decisions:" %}</strong>
-              {% trans "This tool is not a substitute for professional legal advice. Avoid using it for decisions with legal implications." %}
+              {% trans "This tool is not a substitute for professional legal advice. Avoid using it for decisions with legal implications or decisions that would impact an individual." %}
+            </p>
+            <p class="mb-0">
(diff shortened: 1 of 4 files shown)


django/chat/templates/chat/components/chat_welcome.html
```diff
@@ -28,9 +28,12 @@ <h5 class="text-align-center mb-3">{% trans "Safe use of the AI Assistant" %}</h
               <strong>{% trans "No sensitive information" %}:</strong>
               {% trans "Do not upload or process sensitive, confidential, or legally protected documents. This includes personally identifiable information, protected B information, and information protected under solicitor client privilege." %}
             </p>
-            <p class="mb-0">
+            <p>
```
juleskuehn, line 31: Ideally these changes wouldn't be in this PR, but its OK.

django/otto/views.py
```diff
@@ -707,13 +707,13 @@ def cost_dashboard(request):
     # Filter by dates
     if date_group == "last_90_days":
         start_date = timezone.now().date() - timedelta(days=90)
```
juleskuehn, line 709: So that the total time is actually 90, the start date should also be advanced 1 day.
(and same for the others)

django/otto/templates/cost_dashboard.html
```diff
@@ -201,7 +201,7 @@ <h5 class="card-title visually-hidden">{% trans "Chart options" %}</h5>
           {# Bar chart #}
           <div id="chart-container">
             <canvas id="costChart"
-                    width="1526"
+                    width="1530"
```
juleskuehn, line 204: Amazing this was the actual fix. Could you just add a comment under `{# Bar chart #}` above that explains ratios too close to 2:1 will actually cause bugs?

django/otto/views.py
amusbag: Updated to reflect the accurate duration. Oversight on my part. Thank you.

juleskuehn (approved):
Looks good, thanks Ali!

PR #176, opened 2024-12-19
fix: file upload encoding issues

haitmo:
- fixes issue where uploading html documents with Windows-1252 characters fails.
- fixes issue where some docx documents failed.

### DIFF
--- a/django/librarian/utils/process_engine.py
+++ b/django/librarian/utils/process_engine.py
@@ -126,6 +126,7 @@ def guess_content_type(
         "application/csv",
         "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet",
         "officedocument.spreadsheetml.sheet",
+        "application/vnd.openxmlformats-officedocument.wordprocessingml.document",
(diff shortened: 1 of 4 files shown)


django/librarian/utils/process_engine.py
```diff
@@ -126,6 +126,7 @@ def guess_content_type(
         "application/csv",
         "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet",
         "officedocument.spreadsheetml.sheet",
+        "application/vnd.openxmlformats-officedocument.wordprocessingml.document",
         "officedocument.wordprocessingml.document",
         "officedocument.presentationml.presentation",
```
juleskuehn, line 131: I'd say there's a decent chance this same issue will emerge with PPTX - could we add the similar variant for presentation here too?

juleskuehn (approved):
Great! All working on my end. Nice type hints, docstring and test cases too!
I added a comment but it is not a must.

PR #178, opened 2024-12-19
fix: unnecessary preset load operation when saving; bug when user submits incorrect form

mcusteau:
- An unnecessary loading operation when users save a preset is removed

- Fixed bug where modal displays incorrectly in Otto after a user enters incorrect information in the preset form (such as not including a title) , caused by hx-target="#options-accordion"


django/chat/views.py
```diff
@@ -653,18 +653,24 @@ def chat_options(request, chat_id, action=None, preset_id=None):
...
+                    const modalElement = document.getElementById("presets-modal");
+                    let bootstrapModal = bootstrap.Modal.getInstance(modalElement);
+
+                    // Hide the modal
+                    bootstrapModal.hide();
+
+                    // Remove the modal backdrop
+                    document.querySelectorAll('.modal-backdrop').forEach(backdrop => backdrop.remove());
+                    }
+                    </script>
```
mcusteau, line 669: Im removing the modal manually with javascript instead of using data-bs-dismiss="modal" since in the case that the user inserted incorrect information, we want the modal to stay up in order for the user to fix the information in the form

juleskuehn: I'm surprised you have to manually remove the backdrop - are you sure it doesn't work without this?

juleskuehn (commented):
Nice work finding that other bug!

I'd rather just see these save & update routes return 200 rather than the JS to close the modal though.
Can we try going the other way - close the modal using the default bootstrap close handler unless the form *doesn't validate*.

i.e. you keep the data-bs-dismiss, but instead edit your onclick handler so that it calls a validation function? i.e.
```
<button
    data-bs-dismiss="modal"
    ...
    onclick="handlePresetModalSubmit()"
    >
...
<script>
function handlePresetModalSubmit() {
    if (!document.getElementById("#presetModalForm").checkValidity()) { // not sure of the selector actually
      event.preventDefault();
      event.stopPropagation();
      return False;
    }
    document.querySelector('input[name=\'prompt\']').value = document.getElementById('chat-prompt').value;
}
<script>
```

django/chat/templates/chat/modals/presets/presets_form.html
```diff
@@ -1,34 +1,40 @@
...
-                class="nav-link"
-                data-bs-target="#fr-preset-tab-pane"
-                data-bs-toggle="pill"
-                type="button"
-                role="tab"
-                aria-controls="fr-preset-tab-pane"
-                aria-selected="false">{% trans "French" %}</button>
-      </li>
-    </ul>
+  <form class="g-3 mt-1 row"
```
juleskuehn, line 5: Good - we should definitely have the `form` element here.

django/chat/templates/chat/modals/presets/presets_form.html
```diff
@@ -1,34 +1,40 @@
 {% load i18n %}
 
+{% csrf_token %}
```
juleskuehn, line 3: I think this is in 2 places. You can remove it from here, or the other place (line 38).

django/chat/templates/chat/modals/presets/presets_form.html
```diff
@@ -82,52 +88,89 @@
...
-          {% if preset_id %}
-            <button id="delete_preset"
-                    class="btn btn-outline-danger w-25 flex-fill"
-                    hx-get="{% url 'chat:chat_options' chat_id=chat_id action='delete_preset' preset_id=preset_id %}"
+    <div class="modal-footer pb-0 pt-1 border-0">
+      <div class="col-12">
+        <div class="row">
+          <div class="col-6">
+            <button id="cancel_preset"
+                    class="btn btn-outline-secondary w-25 flex-fill me-2"
```
juleskuehn, line 101: Buttons should generally not have specific widths, because the text can overflow depending on screen size. E.g. when the window is tiled to one side, this is what happens:

image

I think this is a pre-existing problem but it would be easy to fix in this PR - simply get rid of the `w-*` classes on any buttons.

django/chat/templates/chat/modals/presets/presets_form.html
```diff
@@ -82,52 +88,89 @@
...
+      if (!nameEnElement && !nameFrElement) {
+          errorMessage.innerText = "{% trans 'Please provide a title in either English or French.' %}";
+          errorMessage.classList.remove('d-none');
+          event.preventDefault();
+          event.stopPropagation();
+          isValid = false;
+      }
+
+      console.log(sharingOption)
+      console.log(accessibleToElement)
```
juleskuehn, line 149: Please remove logging statements

django/chat/views.py
```diff
@@ -650,21 +618,10 @@ def chat_options(request, chat_id, action=None, preset_id=None):
 
                 messages.success(
                     request,
-                    _("Preset saved and loaded successfully."),
+                    _("Preset saved successfully."),
```
juleskuehn, line 621: I changed this. Otherwise, everything is working as expected.

juleskuehn (commented):
Looks good and working correctly! A few minor things that should be fixed before merging. Thanks.

django/chat/templates/chat/modals/presets/presets_form.html
```diff
@@ -82,52 +88,89 @@
             </div>
           </div>
         {% endif %}
-        <div id="error-message">
-          {% if error_message %}<div class="alert alert-danger" role="alert">{{ error_message }}</div>{% endif %}
-        </div>
-      </form>
+        <div id="error-message"
```
juleskuehn, line 91: Could you put this at the top of the form, to be consistent with other error messages in the app?

juleskuehn (approved):
Just made a tiny change to focus the error message. (Useful if window is small since it will scroll up to the message - or, if using screen reader)

Looks good!

PR #180, opened 2024-12-27
fix: create library new library not displayed in libraries list

amusbag:
Fixed bug where new created library didn't show up in the list of libraries after creation. The newly created library wasn't been returned as part of the libraries query due to cached permissions. @juleskuehn suggested invalidating cache right before querying the libraries again which solved the problem.


juleskuehn (approved):
Thanks for figuring this out. Looks good.

PR #181, opened 2024-12-27
ci: reorder github actions & ensure final newline in translations.json

juleskuehn:
The localization GitHub action was erasing the pytest status from the PR (checkmark or X disappears). If we run the localization first, this should solve the problem.

Also, manually editing translations.json will add a newline (due to autoformat settings). This will create spurious diffs vs. the automated translation. So, we ensure a final newline in the automated translation also.

**Update:**
@haitmo This order of GitHub actions still isn't working correctly - can you look into this and update this PR with a fix? Perhaps we need one workflow with multiple jobs, rather than the separate workflows for pytest and localization?
(Although the latter solution would need conditionals for if it is the push to main... in which case don't run localization)
If this is too complicated we could also consider reverting the localization action PR.


.github/workflows/localization.yml
```diff
@@ -1,47 +1,70 @@
...
     permissions:
       contents: write
       pull-requests: write
       actions: read
     steps:
     - uses: actions/checkout@v4
       with:
-        ref: ${{ github.head_ref }}
+        ref: ${{ github.event.pull_request.head.ref }}
+        ssh-key: ${{ secrets.COMMIT_KEY }}
```
haitmo, line 34: the GITHUB_TOKEN cannot trigger new workflow runs. An alternative was to user personal access tokens, but I didn't feel too comfortable with that option since they would have access to _all_ repositories.

Instead I opted to use a dedicated SSH keypair and added it to the repository as a deploy key. If we're not comfortable with that either, I suggest we just:

1-  Forget about it all or,
2- Look for custom actions and reach out to the cloud team to add them to the whitelist

.github/workflows/localization.yml
```diff
@@ -1,47 +1,70 @@
...
+
+    - name: Set environment variable "commit-author"
```
haitmo, line 36: Just setting things up so that the localization command is not run a second time.

haitmo (approved):
Talked with Jules and this is ok to merge to main now.

PR #182, opened 2024-12-27
fix : changed logic of resizing

Rifahaziz:
Different logics of image resizing depending on merged being on or off.
When merge is on - images resized to fit A4 dimensions following a new scale and DPI increased to 300 for clearer resolution.
When merge is off - single image resized as per azure requirements and eliminated pasting image over A4.

### DIFF
--- a/django/text_extractor/utils.py
+++ b/django/text_extractor/utils.py
@@ -19,6 +19,8 @@ from reportlab.lib.pagesizes import A4
 from reportlab.pdfgen import canvas
 from structlog import get_logger
 
+from librarian.utils.process_engine import resize_to_azure_requirements
+
 logger = get_logger(__name__)
 
 from otto.models import Cost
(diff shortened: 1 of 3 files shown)


django/text_extractor/utils.py
```diff
@@ -131,11 +128,40 @@ def resize_image_to_a4(img, dpi=150):
...
+        new_width, new_height = width, height
+    # Edge case: insanely wide or tall images. Don't maintain proportions.
+    new_width = max(50, min(new_width, 10000))
+    new_height = max(50, min(new_height, 10000))
+
+    # Resize the image
+    image = image.resize((new_width, new_height), Resampling.LANCZOS)
+    return image
+
+
```
Rifahaziz, line 159: had to take part of the librarian's method because I was reading the image before passing it here

django/text_extractor/utils.py
```diff
@@ -161,9 +187,19 @@ def create_searchable_pdf(input_file, add_header):
...
+                image_pages = [
+                    resize_image_to_a4(image) for image in image_pages_original
+                ]
+        else:
+            with Image.open(temp_path) as img:
+                image_pages_original = ImageSequence.Iterator(img)
+                image_pages = [
+                    resize_to_azure_images(image) for image in image_pages_original
+                ]
+
```
Rifahaziz, line 202: we do require the a4 background for merged cases only

django/text_extractor/utils.py
```diff
@@ -104,24 +102,23 @@ def calculate_start_pages(files):
...
+    scale = min(a4_width / img.width, a4_height / img.height)
 
-    if img_ratio > a4_ratio:
-        # Image is wider than A4 ratio, fit by width
-        new_width = int(a4_width / 2)
-        new_height = int(new_width / img_ratio)
+    if scale < 1.0:
+        # Image is bigger than A4 in at least one dimension; scale it down
+        new_width = int(img.width * scale)
+        new_height = int(img.height * scale)
```
Rifahaziz, line 114: Even when merged, images won't be shortened more than the a4's width or height.

django/text_extractor/utils.py
juleskuehn: You could instead make a small change to the function so that it will support both file and string inputs (i.e. a type check on the input). Then you can import it instead of copy pasting.

juleskuehn (dismissed):
Looks great, works great, ideally you would avoid copy/pasting the function though as I commented. I am not too concerned about it so I am approving anyway. If you have time to make the change though, I would appreciate it and will quickly re-review.

Rifahaziz:
I have modified the librarian method so that it can be used by text extractor too @juleskuehn

juleskuehn (approved):

PR #191, opened 2025-01-03
fix: eliminate unnecessary display functions

smcdonald-jus:
In the course of working on something else, I noticed that there were 500 errors popping up in the Sandbox feedback manager. It turned out to be related to a feedback item with an obsolete Type category, which was breaking the custom feedback_type_display function in the Feedback model.

Fortunately, it turns out that Django provides a built-in display method in HTML templates, which does the same thing as the custom function but also allows for obsolete categories to be displayed.

### DIFF
--- a/django/otto/models.py
+++ b/django/otto/models.py
@@ -338,12 +338,6 @@ class Feedback(models.Model):
 
     objects = FeedbackManager()
 
-    def status_display(self):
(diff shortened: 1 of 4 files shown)


haitmo (approved):
Thanks Shaun, it was on my to-do list, but its nice to see it already done!

PR #192, opened 2025-01-06
fix: qa edit modal library change

amusbag:
Fixed a bug where the user changes the library using the modal but it wouldn't reflect the sources when user prompts the system. Options weren't being saved after the change.


juleskuehn (approved):
Simple fix - thanks! HTMX "then" syntax is nice.

PR #193, opened 2025-01-07
feat: add smooth scrolling to current chat list item and enhance button visibility

haitmo:
Makes sure chat is visible in sidebar onload

Ellipsis is shown by default for current chat as well.

image

### DIFF
--- a/django/chat/static/chat/js/scripts.js
+++ b/django/chat/static/chat/js/scripts.js
@@ -100,7 +100,7 @@ function scrollToListItem() {
     const currentChat = document.querySelector('.chat-list-item.current');
     if (currentChat) {
       currentChat.scrollIntoView({
-        behavior: 'smooth',
+        behavior: 'instant',
         block: 'center'
       });
     }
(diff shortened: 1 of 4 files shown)


django/chat/templates/chat/chat.html
```diff
@@ -27,7 +27,8 @@
                 style="z-index:1"
                 type="button"
                 aria-controls="left-sidebar"
-                id="left-sidebar-toggle">
+                id="left-sidebar-toggle"
+                onClick="scrollToListItem();">
```
juleskuehn, line 31: Thanks for handling the case in which sidebar is not initially shown!

django/chat/static/chat/js/scripts.js
```diff
@@ -95,6 +95,18 @@ function scrollToBottom(smooth = true, force = false) {
   messagesContainer.scrollTop = hashContainer ? messagesContainer.scrollTop + offset : messagesContainer.scrollHeight;
 }
 
+function scrollToListItem() {
+  setTimeout(() => {
+    const currentChat = document.querySelector('.chat-list-item.current');
+    if (currentChat) {
+      currentChat.scrollIntoView({
+        behavior: 'smooth',
```
juleskuehn, line 103: Personally I'd prefer it didn't smooth scroll (a lot going on with 2 panes scrolling at once on load)

juleskuehn (approved):
Looks great. I would remove the behavior: smooth, which I think is too much in this case.
Thanks!

juleskuehn (approved):

PR #194, opened 2025-01-08
fix: qa advanced setting

amusbag:
Fixed a bug where QA advanced settings options weren't being reset even though the QA autocomplete form was being reset. Value of qa mode was being changed but change event wasn't being dispatched which is needed to trigger the actual reset.


juleskuehn (dismissed):
Nice. Can you just confirm this also works when loading a settings preset? (I imagine it does - just check before merging)

juleskuehn (approved):

PR #196, opened 2025-01-15
feat: prompt generator in AI assistant

Rifahaziz:


django/chat/templates/chat/components/chat_input.html
```diff
@@ -86,26 +86,55 @@
...
+                  hx-trigger="click"
+                  hx-indicator="#loading-spinner"
+                  aria-label='{% trans "Improve prompt using AI" %}'
+                  title='{% trans "Improve prompt using AI" %}'
+                  onclick="toggleSpinner(true)"
+                  style="top: 0px;
+                         right: 40px;
+                         width: 30px;
+                         height: 30px;
+                         display: flex;
```
Rifahaziz, line 114: this style does not apply if i add it in style.css. i dont know why this gets overridden

django/chat/views.py
```diff
@@ -907,3 +910,131 @@ def update_qa_options_from_librarian(request, chat_id, library_id):
...
+                "role": "system",
+                "content": META_PROMPT,
+            },
+            {
+                "role": "user",
+                "content": "Task, Goal, or Current Prompt:\n" + task_or_prompt,
+            },
+        ],
+    )
+    llm.create_costs()
```
Rifahaziz, line 1009: could not verify if this is actually adding to the cost. What's the best way to verify this?

juleskuehn: Inspect the Cost objects in django shell.

django/chat/templates/chat/components/chat_input.html
juleskuehn: If you are targeting `#magic-prompt` in chat/static/chat/style.css, I don't see why it wouldn't?
You may have to Ctrl+Shift+R to see changes in the browser.

Rifahaziz:
The changes with modal will be in the old branch "feat-prompt-generator"

PR #202, opened 2025-01-21
fix: edit library modal user perm

amusbag:
Fixed a bug where AI Assistant Users are redirected to "index" in edit library modal because they do not have permission to edit the "Corporate" library.


django/otto/utils/decorators.py
```diff
@@ -85,9 +85,14 @@ def _wrapped_view(request, *args, **kwargs):
                 # User does not have a required permission
                 if raise_exception:
                     raise PermissionDenied()
+                # User does not have required permission to edit library so redirect to their personal library
+                elif perms.__contains__("librarian.edit_library"):
+                    return redirect(
+                        f"/librarian/modal/library/{request.user.personal_library.id}/edit"
```
juleskuehn, line 91: Can you use the Django "reverse" function to generate the URL?

juleskuehn (commented):
Just one change to make URL resilient to upstream changes in urls.py etc; I think this is good to go otherwise.

juleskuehn (dismissed):
Looks good and working on my end. Thanks!

juleskuehn (approved):

PR #203, opened 2025-01-21
feat: prompt generator in chat

Rifahaziz:
Prompt generation using AI to assist users in prompt engineering. The magic icon for prompt generation is available for chat mode only. The feature includes a modal to open on clicking the icon and at the back-end openai's prompt generator api is used. Users can regenerate and edit their prompts and costs to each api call is shown.


django/chat/static/chat/js/scripts.js
```diff
@@ -108,6 +108,10 @@ function scrollToListItem() {
 }
 
 function handleModeChange(mode, element = null, preset_loaded = false) {
+  console.log(`mode::::::::::::: ${mode}`);
+  if (!mode || mode === '') {
+    mode = 'chat';
+  }
   // Set the hidden input value to the selected mode
```
Rifahaziz, line 115: one issue here : when we go to any AI assistants from the home page, nothing is stored in the 'mode' variable. This variable is only updated when we switch the chat/qa/summarize/translate "tabs" and not when we select them from the home page.

django/chat/templates/chat/components/chat_input.html
```diff
@@ -94,6 +94,25 @@
...
+                  data-bs-target="#magicModal"
+                  hx-get="/reset-form/"
+                  hx-include="#chat-prompt"
+                  hx-target="#magicModal .modal-content"
+                  hx-swap="outerHTML"
+                  aria-label='{% trans "Help me write a good prompt" %}'
+                  title='{% trans "Help me write a good prompt" %}'
+                  onclick="setMagicPrompt()"
+                  style="background-color: transparent;
+                         z-index: 5">
```
Rifahaziz, line 111: all the other style variables are applied in css file, except the z-index and background color. These 2 gets overridden when shifted to style.css file.

django/chat/views.py
```diff
@@ -910,3 +913,132 @@ def update_qa_options_from_librarian(request, chat_id, library_id):
...
+            output_text, cost = generate_prompt(user_input)
+            output_text = re.sub(
+                r"<reasoning>.*?</reasoning>", "", output_text, flags=re.DOTALL
+            )
+
+            return render(
+                request,
+                "chat/components/generated_prompt.html",
+                {"user_input": user_input, "output_text": output_text, "cost": cost},
+            )
```
Rifahaziz, line 1035: costs displayed

django/chat/static/chat/js/chatInput.js
```diff
@@ -83,3 +83,73 @@ document.addEventListener('mouseup', function () {
...
+      if (data.output_text) {
+        document.getElementById('generated-prompt').value = data.output_text;
+      } else {
+        console.error('Error:', data.error);
+      }
+    })
+    .catch(error => {
+      console.error('Error:', error);
+    });
+});
```
juleskuehn, line 114: Can you do this with HTMX?

django/chat/static/chat/js/chatInput.js
```diff
@@ -83,3 +83,73 @@ document.addEventListener('mouseup', function () {
...
+  const magicIcon = document.getElementById('generate-text');
+  const spinnerIcon = document.getElementById('spinner-icon');
+  if (show) {
+    magicIcon.style.display = 'none';
+    spinnerIcon.style.display = 'inline-block';
+  } else {
+    magicIcon.style.display = 'inline-block';
+    spinnerIcon.style.display = 'none';
+  }
+}
```
juleskuehn, line 127: You should also be able to do this with HTMX (hx-indicator) - the indicator will just be outside the element. I think everything in this file could be done with HTMX to reduce complexity. Please let me know if I'm missing something though.

django/chat/static/chat/js/scripts.js
juleskuehn: Interesting - I will look into that!

django/chat/static/chat/style.css
```diff
@@ -1271,3 +1271,28 @@ math[display="block"] {
...
+.wand-button {
+  background-color: transparent;
+  border: none;
+  top: 0px;
+  right: 40px;
+  width: 35px;
+  height: 30px;
+  display: flex;
+  position: absolute;
+  z-index: 5">
```
juleskuehn, line 1284: Formatting error here probably from copy/paste. This is likely why the z-index wasn't working here.

django/chat/static/chat/style.css
```diff
@@ -1271,3 +1271,28 @@ math[display="block"] {
   margin-top: 1rem;
   padding-bottom: 1rem;
 }
+
+.wand-button {
+  background-color: transparent;
```
juleskuehn, line 1276: You can probably make this style rule work like this: `background-color: transparent !important;`

django/chat/templates/chat/components/chat_input.html
```diff
@@ -105,7 +124,59 @@
...
+                 class="form-control "
+                 id="magic-prompt"
+                 placeholder='{% trans "Type your prompt here..." %}'>
 
+          <button type="submit"
+                  class="btn btn-primary mt-3"
+                  id="generate-prompt"
+                  hx-post="/chat/generate-prompt/"
+                  hx-trigger="click"
+                  hx-indicator="#loading-spinner"
```
juleskuehn, line 167: This is why I'm confused about some of the Javascript - I would think the HTMX would be enough.

django/chat/templates/chat/components/generated_prompt.html
```diff
@@ -0,0 +1,43 @@
...
+
+<div class="modal-body">
+
+  <form id="magic-form col-md-4"
+        hx-post="/chat/generate-prompt/"
+        hx-target="#magicModal .modal-body"
+        hx-swap="outerHTML">
+    {% csrf_token %}
+ 
+    <h6>Your prompt</h6>
```
juleskuehn, line 12: Add translation tag

django/chat/templates/chat/components/chat_input.html
```diff
@@ -105,7 +124,59 @@
...
+          <input type="text"
+                 name="user_input"
+                 class="form-control "
+                 id="magic-prompt"
+                 placeholder='{% trans "Type your prompt here..." %}'>
 
+          <button type="submit"
+                  class="btn btn-primary mt-3"
+                  id="generate-prompt"
+                  hx-post="/chat/generate-prompt/"
```
juleskuehn, line 165: Don't hardcode URLs; use the Django template URL tag. This feedback applies to all URLs in the code.

django/chat/templates/chat/components/generated_prompt.html
```diff
@@ -0,0 +1,43 @@
...
+    <textarea class="form-control mt-3"
+              id="generated-prompt"
+              name="generated_prompt"
+              rows="10">{{ output_text }}</textarea>
+    <p>{% trans "Cost : " %} {{ cost }}</p>
+  </form>
+  <div class="modal-footer">
+    <button type="button" class="btn btn-secondary"  onclick="clearTextAreas()">{% trans "Clear" %}</button>
+    <button type="button" class="btn btn-secondary"  onclick="location.reload();">{% trans "Reset" %}</button>
+
```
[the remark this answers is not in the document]
juleskuehn: I don't think you should need to refresh the page to clear the state. I'll test it out and see though!

django/chat/templates/chat/components/generated_prompt.html
```diff
@@ -0,0 +1,43 @@
...
+              rows="10">{{ output_text }}</textarea>
+    <p>{% trans "Cost : " %} {{ cost }}</p>
+  </form>
+  <div class="modal-footer">
+    <button type="button" class="btn btn-secondary"  onclick="clearTextAreas()">{% trans "Clear" %}</button>
+    <button type="button" class="btn btn-secondary"  onclick="location.reload();">{% trans "Reset" %}</button>
+ 
+    <button type="button"
+            class="btn btn-primary"
+            onclick="applyGeneratedPrompt(event)">{% trans "Use this prompt" %}</button>
```
juleskuehn, line 41: This is something we should definitely use vanilla JS for rather than HTMX. Just not the POST or GET stuff.

django/chat/views.py
```diff
@@ -910,3 +917,120 @@ def update_qa_options_from_librarian(request, chat_id, library_id):
...
         },
     )
+
+
+def generate_prompt(task_or_prompt: str):
+
+    load_dotenv()
+    openai.api_key = os.getenv("AZURE_OPENAI_KEY")
+
+    openai.api_version = os.getenv("AZURE_OPENAI_VERSION")
```
juleskuehn, line 927: You shouldn't need these 3 lines now, or the imports for them, since you are using OttoLLM class.

django/chat/views.py
```diff
@@ -910,3 +917,120 @@ def update_qa_options_from_librarian(request, chat_id, library_id):
             "trigger_library_change": "true" if library != original_library else None,
         },
     )
+
+
+def generate_prompt(task_or_prompt: str):
```
juleskuehn, line 922: This can be moved to utils.py - trying to keep views.py only for functions that handle requests.

django/chat/urls.py
```diff
@@ -110,4 +110,6 @@
         views.set_preset_default,
         name="set_preset_default",
     ),
+    path("reset-form/", views.reset_form_view, name="reset_form_view"),
```
juleskuehn, line 113: I don't think this is used anywhere.

django/chat/views.py
```diff
@@ -910,3 +917,120 @@ def update_qa_options_from_librarian(request, chat_id, library_id):
...
+                "content": "Task, Goal, or Current Prompt:\n" + task_or_prompt,
+            },
+        ],
+    )
+    usd_cost = llm.create_costs()
+    cost = display_cad_cost(usd_cost)
+    return completion.choices[0].message.content, cost
+
+
+def reset_form_view(request):
```
juleskuehn, line 1013: This isn't used anywhere.

django/chat/views.py
```diff
@@ -910,3 +917,120 @@ def update_qa_options_from_librarian(request, chat_id, library_id):
...
+    [Optional: 1-3 well-defined examples with placeholders if necessary. Clearly mark where examples start and end, and what the input and output are. User placeholders as necessary.]
+    [If the examples are shorter than what a realistic example is expected to be, make a reference with () explaining how real examples should be longer / shorter / different. AND USE PLACEHOLDERS! ]
+
+    # Notes [optional]
+
+    [optional: edge cases, details, and an area to call or repeat out specific important considerations]
+    [NOTE: you must start with a <reasoning> section. the immediate next token you produce should be <reasoning>]
+    """.strip()
+
+    completion = openai.chat.completions.create(
```
juleskuehn, line 995: You can't use this because it won't tie into the cost tracking. You have to use LlamaIndex interface to OpenAI models, through the OttoLLM class.

juleskuehn (commented):
Thanks Rifah, this is a fairy big feature and from your demo of it, I think you've done a good job on thinking through and implementing the user interface.

Most of the comments aim to make the code simples. The one about cost tracking is most important to address. I'll test out the functionality locally too and follow up with more changes. We could work together on the HTMX refactoring if it is behaving strangely for you.

django/chat/templates/chat/components/generated_prompt.html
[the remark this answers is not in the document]
Rifahaziz: should I just get rid of reset?

django/chat/static/chat/js/chatInput.js
Rifahaziz: I was in the beginning, but I was having issues with the spinner appearing. Spinner with htmx is still not working.

juleskuehn (approved):

PR #206, opened 2025-01-24
feat: terms of use page

juleskuehn:
**Note to reviewers:** Please set `ENV=PROD` in your .env and restart Django to see the intended messages for the Pilot environment.

* Change warnings etc. to be more succinct and reflect protected B ATO (in "prod" environment only)
* Add a persistent link to the terms of use from user dropdown menu and homepage.
  * If not accepted, contains a single checkbox to accept
  * If already accepted, form is hidden and a message "You agreed to these terms on X date" is shown.
* Write a new "terms of use" page that conglomerates all relevant terms of use in more detail and links to policies.
* Conditionally display "Protected B approved" messages depending on environment


django/chat/templates/chat/components/chat_welcome.html
```diff
@@ -20,57 +20,66 @@ <h5 class="text-align-center mb-3">{% trans "Safe use of the AI Assistant" %}</h
...
           <div class="col">
+            {% if environment == "prod" %}
+              <p>
+                <strong>{% trans "Protected B approved:" %}</strong>
+                {% trans "Use with unclassified, Protected A or Protected B information only." %}
+              </p>
+            {% else %}
+              <p>
+                <strong>{% trans "Unclassified only:" %}</strong>
+                {% trans "This is a development / testing environment. Use unclassified information only." %}
```
smcdonald-jus, line 31: Pro-A isn't allowed in non-prod either?

smcdonald-jus (approved):

PR #210, opened 2025-01-28
fix: increase toast duration and added toasts

haitmo:
Increased toasts duration from 2s to 5s.

Increased z-index of toasts to appear before modals(z: 1055)

image


django/otto/static/style.css
```diff
@@ -388,3 +388,8 @@ ul.errorlist ul:not(:last-child) {
 ul.errorlist p {
   margin-bottom: 0.25rem;
 }
+
+.toast-sticky {
+  /* increase z-index to show over modals */
+  z-index: 1070;
```
haitmo, line 394: Modals have a  z-index value of 1055

django/otto/templates/components/messages.html
```diff
@@ -1,5 +1,5 @@
 {% load i18n %}
-{% if messages %}
+{% if messages and has_error %}
```
haitmo, line 2: As discussed, only error messages are permanently displayed in the librarian modal. Success messages appear in toasts.

juleskuehn (approved):
5 seconds feels a bit long but I'm fine with it. Working well, looks good!

PR #217, opened 2025-01-30
fix: cancel button for creating new library and data source fixed

Rifahaziz:


django/librarian/views.py
```diff
@@ -327,6 +327,10 @@ def modal_library_list(request):
     return modal_view(request)
 
 
+def modal_datasource_list(request, library_id):
```
juleskuehn, line 330: Could you just use modal_edit_library, and avoid creating this view, URLs etc?

django/librarian/templates/librarian/forms/data_source.html
```diff
@@ -7,7 +7,7 @@
   {% url 'librarian:modal_delete_data_source' data_source_id=form.instance.id as delete_url %}
 {% else %}
   {% url 'librarian:modal_create_data_source' library_id=form.library.value as form_action_url %}
-  {% url 'librarian:modal_library_list' as cancel_url %}
+  {% url 'librarian:modal_datasource_list' library_id=form.library.value as cancel_url %}
```
juleskuehn, line 10: Ah, thanks for finding this. I think you could just use `librarian:modal_edit_library` instead though.

django/librarian/views.py
Rifahaziz: yes, done.

juleskuehn (approved):
Looks good! Thanks for renaming "modal" URL so it is more clear.

Rifahaziz:
> Looks good! Thanks for renaming "modal" URL so it is more clear.

Thanks! It was already renamed to modal_library_list, but it was getting called by modal. The url name was a mismatch.

PR #229, opened 2025-02-03
fix: language check in terms of use

juleskuehn:
For an unknown-to-me reason, the variable *has to be* named LANGUAGE_CODE for this check to work.


amusbag (approved):
Looks good. LANGUAGE_CODE because of the global language (LANGUAGE_CODE = 'en-ca') in settings.py maybe?

PR #232, opened 2025-02-04
fix: streaming issue in test_chat_views file

Rifahaziz:
Seems like the problem was using final_response(); it was not storing the error_id , rather it stored the location of the content.
requirements.txt is changed, so please install the new updates before running.


juleskuehn (approved):
Thanks!

PR #234, opened 2025-02-06
fix: change template wizard description and tile order on index page

amusbag:
- Changed description_en and description_fr of Template Wizard in app.yaml
- Create empty migration and ran reset_apps command
- Sorted the category_features by title which as a temporary solution to reordering so as to not have to make changes to CATEGORY_CHOICES in models since this is temporary while Template Wizard is under development


juleskuehn:
Let's just forget about the resorting. Sorting stuff alphabetically is not recommended because tiles will jump around English vs French, right?

amusbag:
> Let's just forget about the resorting. Sorting stuff alphabetically is not recommended because tiles will jump around English vs French, right?

No, I tested it and it's sorted the properly in French as well.

juleskuehn (approved):

PR #236, opened 2025-02-06
fix: accessible_to search bar appears when private option is selected

mcusteau:
https://dev.azure.com/BusinessAnalyticsCentre/BAC%20Agile/_workitems/edit/3129

Fixes instances where the search bar to select users for the 'other' preset sharing option appears in the modal even if 'private' is selected. I also fixed some confusing variable names in the javascript code.


juleskuehn (approved):
Nice catch. Thanks

PR #239, opened 2025-02-10
perf(laws): jsonb metadata index, hnsw params

juleskuehn:
- Use JSONB for storing metadata for laws and add GIN index on metadata
  - reduced from **2.6s to 0.07s per iteration**
- Build HNSW index *after* loading all laws.
  - increases accuracy and performance (see https://www.pinecone.io/blog/hnsw-not-enough/)
  - improves speed of loading laws
- Adjust HNSW parameters
  - TODO: grid search / benchmark different parameters for laws search specifically.
  - Based on benchmarks here: https://www.pinecone.io/learn/series/faiss/hnsw/

Need to manually run `load_laws_xml --reset` when upgrading. For local testing you could do this with the "--small" flag.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,23 +127,7 @@ python django/manage.py load_laws_xml --reset
 * To load all laws (slow and quite expensive - around $20; 8 hours), add the `--full` flag.
 * If you leave off `--reset` it should only add laws which aren't already loaded, so you can incrementally add more.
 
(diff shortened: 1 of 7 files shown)


amusbag (approved):
Looks good. All checks passed.

smcdonald-jus (approved):

PR #242, opened 2025-02-11
feat: add markdown to sources

mcusteau:
https://dev.azure.com/BusinessAnalyticsCentre/BAC%20Agile/_workitems/edit/3124

THIS TICKET COVERS:

- Javascript code to render sources in markdown when user clicks on the accordion buttons
- python code that creates clickable internal links for the sources
- The source url at the top right is now a button to differentiate it from links in the source text
- css code to set to font size to medium so markdown doesnt look crazy with big headings
- css code to stylize the accordion colors so they are not ugly


juleskuehn:
Almost there - I would like to see the source markdown passed through in a data-md attribute though (like in chat responses). See in chat/utils.py:
```
def wrap_llm_response(llm_response_str):
    return f'<div class="markdown-text" data-md="{html.escape(json.dumps(str(llm_response_str)))}"></div>'
```

juleskuehn (approved):
Everything looking good now - thanks for adding the tests and addressing feedback!

PR #243, opened 2025-02-12
fix: law sources view needs json.loads (after switch to JSONB)

juleskuehn:
Added Pytest also.

To test manually, search for something in the laws view then click the 3 dots on one of the returned sources. (in main branch, this causes an error)


amusbag (approved):
Looks good. Reproduced the error manually in main, error didn't occur in this branch. Test passed manually.

PR #245, opened 2025-02-12
perf: use jsonb for all llama_index metadata

juleskuehn:
You can test locally by just running `python manage.py migrate`, ensuring there are no errors, and then querying some QA libraries.

Note that this does not create metadata indexes. That is still only on the laws table. Query performance on the normal QA libraries would not be improved much if at all from the indexes.


amusbag (approved):
Looks good.

PR #247, opened 2025-02-13
ops: goc root cert revisions

jannable:
- Added GoC root cert fetch and install to Dockerfle
- Fixed reference to requirements_dev.txt
- Repaired vectordb port to account for pgbouncer
- Removed llama_index creation because it's in k8s yaml now (had troubles getting that to work through pgbouncer)


django/otto/management/commands/reset_app_data.py
```diff
@@ -37,33 +37,11 @@ def handle(self, *args, **options):
         vector_db_user = settings.DATABASES["vector_db"]["USER"]
```
haitmo, line 37: With the exception of the password, can we not remove all the other variables now then?

haitmo (approved):

PR #254, opened 2025-02-17
fix drag leave events

juleskuehn:
Fixes issue where dragging files into the AI assistant - but not completing the drag-drop action - would keep the dropzone lightbox on the screen indefinitely.
Now the dropzone disappears when dragging cursor leaves the window.


amusbag (approved):
Replicated the issue in main and it's now fixed in PR.

PR #256, opened 2025-02-17
fix: restore vector db creation code, make migrations, index corp library in initial_setup

smcdonald-jus:
Restore the code in post-create.py that creates the vector db, but only for local environments. Also made some migration files that weren't included after some recent model changes. Finally, initial_setup now starts (and subsequently terminates) a Celery worker to index corporate library docs upon loading them.


django/laws/management/commands/load_laws_xml.py
```diff
@@ -863,7 +863,7 @@ def handle(self, *args, **options):
             "DROP INDEX IF EXISTS laws_lois_node_id__idx;"
             "CREATE INDEX laws_lois_node_id__idx ON data_laws_lois__ (node_id);"
             "DROP INDEX IF EXISTS laws_lois_metadata__idx;"
-            "CREATE INDEX laws_metadata_index ON data_laws_lois__ USING GIN (metadata_);"
+            "CREATE INDEX laws_lois_metadata__idx ON data_laws_lois__ USING GIN (metadata_);"
```
juleskuehn, line 866: thank you

juleskuehn (approved):
Thanks, Shaun!

mcusteau (approved):

haitmo (approved):

PR #257, opened 2025-02-18
feat: highlight pdf in-text sources

Rifahaziz:
claims done by the llm in its references are highlighted in the sources modal


django/chat/models.py
```diff
@@ -515,6 +515,8 @@ class AnswerSource(models.Model):
     min_page = models.IntegerField(null=True)
     max_page = models.IntegerField(null=True)
 
+    claims_list = models.JSONField(default=list, blank=True)
```
juleskuehn, line 518: Add nullable "highlighted_text" as a property here also, then you only have to highlight it once (subsequent clicks of 'sources' will not run it again). Update the `node_text` property so that it will use highlighted_text if it exists.

django/chat/utils.py
```diff
@@ -194,10 +195,13 @@ def sse_string(
         sse_joiner = "\ndata: "
         if wrap_markdown:
             message = wrap_llm_response(message)
+
         if dots:
             message += dots
-        out_string = "data: "
+
+        out_string = "data: " + message
```
juleskuehn, line 202: I think all changes to this file should be reverted.

django/chat/views.py
```diff
@@ -706,20 +707,79 @@ def set_security_label(request, chat_id, security_label_id):
...
+                good_matches.append(chunk_text)
+
+    for match in good_matches:
+        if len(match) > 3:
+            text = text.replace(match, f"<mark>{match}</mark>")
+
+    return text
+
+
+from chat.llm import OttoLLM
```
juleskuehn, line 741: move to the top of the file please

django/chat/views.py
```diff
@@ -706,20 +707,79 @@ def set_security_label(request, chat_id, security_label_id):
...
+    etc.
+
+    ---
+    <llm_response>
+    {llm_response_text}
+    </llm_response>
+    """
+    claims_response = llm.complete(prompt)
+    # find the claim tags and add whats wrapped in the claim tags to a list
+    claims_list = re.findall(r"<claim>(.*?)</claim>", claims_response)
```
juleskuehn, line 762: Create costs before returning

django/chat/views.py
```diff
@@ -706,20 +707,79 @@ def set_security_label(request, chat_id, security_label_id):
...
+
         source_text = str(source.node_text)
 
         def replace_page_tags(match):
             page_number = match.group(1)
             return f"<span class='fw-semibold'>Page {page_number}</span>"
 
         modified_text = re.sub(r"<page_(\d+)>", replace_page_tags, source_text)
+        claims_list = source.claims_list
+        modified_text = highlight_claims(claims_list, modified_text)
```
juleskuehn, line 782: You could potentially make all the changes you need in models.py - then no need to do any fiddling here.

django/chat/views.py
```diff
@@ -706,20 +707,79 @@ def set_security_label(request, chat_id, security_label_id):
...
+
 @permission_required("chat.access_message", objectgetter(Message, "message_id"))
 def message_sources(request, message_id):
     sources = []
 
     for source in AnswerSource.objects.prefetch_related(
         "document", "document__data_source", "document__data_source__library"
     ).filter(message_id=message_id):
+
         source_text = str(source.node_text)
```
juleskuehn, line 774: Because `source.node_text` is actually a function, you can update that function to do the claim highlighting (if no highlighted_text property) or just return the pre-highlighted text.

django/chat/utils.py
```diff
@@ -276,7 +276,10 @@ def sse_string(
                 )
             await asyncio.sleep(0.01)
 
-        yield sse_string(full_message, wrap_markdown, dots=False, remove_stop=True)
+        yield sse_string(
+            full_message, wrap_markdown=False, dots=False, remove_stop=True
```
Rifahaziz, line 280: copied from main branch

django/chat/utils.py
Rifahaziz: done, copied from main branch

django/chat/views.py
Rifahaziz: do we need to show this cost in the sources modal? or just keeping its track in the background is enough?

juleskuehn: No to the modal. I think what makes most sense is to do this LLM call at the end of the chat response (from `htmx_stream`), passing in the LLM being used in the response - then it will get added to the chat message cost before the cost is shown in the message.

juleskuehn (approved):
Working well now - needs more tests, please make a follow up PR for that. Thanks!

juleskuehn (approved):

PR #258, opened 2025-02-19
fix: show cost tooltips. Improve French localization for user cost display

haitmo:
Fixed the cost tooltip to show on hover.
Adjusted French text.


django/otto/templates/components/user_cost.html
```diff
@@ -12,16 +12,16 @@
...
-            id="cost-bar-inner"
-            class="w-100 d-block"
-            style="height: 12px"
-            min="0"
-            max="100"
-            low="60"
-            high="80"
-            optimum="0"
-            value="{{ cost_percent }}">{{ cost_tooltip }}</meter>
+       <meter aria-label="{{ cost_label }}"
```
juleskuehn, line 15: Weird - I just edited this file and I think my formatter actually un-indented this section.

PR #270, opened 2025-03-03
feat: msg file email loading

amusbag:
Load contents of msg files (i.e. from, to, body, subject, attachments, etc.)
If the msg file contains another msg file as an attachment, then load the contents of that as well.


juleskuehn (dismissed):
Still some very minor bugs with nested file_paths, but I created a ticket to address that later.

juleskuehn (approved):

PR #272, opened 2025-03-05
feat: date chat history

mcusteau:
The javascript code is for deleting the empty sections after a user deletes the last chat in it


django/chat/migrations/0005_chat_last_message_date.py
```diff
@@ -0,0 +1,19 @@
...
+
+    dependencies = [
+        ("chat", "0004_alter_message_options"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="chat",
+            name="last_message_date",
+            field=models.DateTimeField(default=django.utils.timezone.now),
```
juleskuehn, line 24: Should set this to the chat.last_accessed in the migration. I will push a commit that does this.

django/chat/models.py
```diff
@@ -69,6 +69,8 @@ class Chat(models.Model):
     # Last access time manually updated when chat is opened
     accessed_at = models.DateTimeField(auto_now_add=True)
 
+    last_message_date = models.DateTimeField(default=timezone.now)
```
juleskuehn, line 72: I don't really think you need this. You could do a query on Message model to get it. Try that and check the query performance in the Django debug toolbar. If performance with the Message query is fine you can just get rid of all changes to this file, the migration, and the merge migration. I'm fine with it either way though - just seems redundant.

django/chat/static/chat/style.css
```diff
@@ -415,6 +415,11 @@ html[lang="fr"] .chat-sidebar {
   display: none;
 }
 
+.chat-history-section-header {
+  padding: 10px 0 5px;
+  font-weight: 500;
```
juleskuehn: You could accomplish this with bootstrap classes (specifically, `fw-semibold`) but I actually suggest a thinner and smaller muted style - will push

django/chat/templates/chat/components/chat_list_item_menu.html
```diff
@@ -18,7 +18,7 @@
     <h6 class="dropdown-header" role="presentation">{% trans 'Options' %}</h6>
     <div class="list-group list-group-flush options-menu">
       <button type="button"
-              hx-get="{% url 'chat:rename_chat' chat_id=chat.id current_chat=chat.current_chat %}"
+              hx-get="{% url 'chat:rename_chat' chat_id=chat.id section=chat.section current_chat=chat.current_chat %}"
```
juleskuehn, line 21: I don't really understand why you need to keep track of the chat section numbers?

django/tests/chat/test_chat_procs.py
```diff
@@ -424,3 +427,46 @@ async def stream_generator3():
     assert await pruning_test_stream.__anext__() == "<|batchboundary|>"
     # Finally, return pruning message due to empty final stream
     assert await pruning_test_stream.__anext__() == _("**No relevant sources found.**")
+
+
+@pytest.mark.django_db
+def test_get_chat_history_sections(client, all_apps_user):
```
juleskuehn, line 433: I was refactoring this function when you pushed more changes. Here are the changes I made. Just reduce repeated function calls, make it cleaner and translate the labels.

```python
def get_chat_history_sections(user_chats):
    chat_history_sections = [
        {"name": _("Today"), "chats": []},
        {"name": _("Yesterday"), "chats": []},
        {"name": _("Last 7 days"), "chats": []},
        {"name": _("Last 30 days"), "chats": []},
        {"name": _("Older"), "chats": []},
    ]

    today = timezone.now().date()
    for user_chat in user_chats:
        last_message_date = user_chat.last_message_date.date()
        if last_message_date > today - timezone.timedelta(days=1):
            chat_history_sections[0]["chats"].append(user_chat)
        elif last_message_date > today - timezone.timedelta(days=2):
            chat_history_sections[1]["chats"].append(user_chat)
        elif last_message_date > today - timezone.timedelta(days=7):
            chat_history_sections[2]["chats"].append(user_chat)
        elif last_message_date > today - timezone.timedelta(days=30):
            chat_history_sections[3]["chats"].append(user_chat)
        else:
            chat_history_sections[4]["chats"].append(user_chat)

    return chat_history_sections
```

juleskuehn (commented):
It was working when I tested it but there have been changes since. I assume it still works! Good job, it looks great.

django/tests/chat/test_chat_procs.py
juleskuehn: Hm, this comment showed up in the wrong place - mean in chat.utils

juleskuehn (dismissed):
Looks good to me!

PR #274, opened 2025-03-07
feat: making prompt icon more noticeable

Rifahaziz:
-removed redundant settings button
-magic icon with label moved in old settings' position
-prompt generator visible for chat and qa mode only


juleskuehn (approved):
made a few tweaks such as focusing the field when modal opens and removing some custom styles.

i suggest we don't show it in Q&A mode until we get the query rewriting implemented - the retrieval performance would be really bad. if you and abe think its better to show it anyway, that's fine - i'd just try to get the query rewriting implemented in the next week then.

PR #279, opened 2025-03-12
fix: add hx-redirect to login when required

juleskuehn:
To replicate existing issue:
* on `main` branch, open Otto. Log in. Open a second tab. Log out in that tab. Return to the first Otto tab. After 1 minute, the HTMX request to user_cost will result in the login page being superimposed over the homepage.

To test this branch:
* repeat the same steps. The first tab should redirect to the login page rather than superimposing it.


smcdonald-jus:
Yep, it works

smcdonald-jus (approved):

PR #282, opened 2025-03-14
impr: q&a default prompt - reduce hallucinations

juleskuehn:
Improves adherence to "context-only" rule.

Unfortunately I *cannot* get the data migration (trigger reset_app_data) to work. So this command will have to be run manually in sandbox shell.


smcdonald-jus (dismissed):
The amount of improvement facilitated by this change is roughly the size of a barge.

django/chat/templates/chat/modals/advanced_qa_modal.html
```diff
@@ -130,7 +130,7 @@ <h5 class="modal-title" id="advanced-qa-modal-title">
           <div class="col-lg-4 mb-2">
             <label for="qa_system_prompt-modal" class="col-form-label-sm pb-0">{% trans "System prompt" %}</label>
             <p class="text-secondary small mb-2">
-              <em>{% trans "(Optional) Define AI model's role, background information, rules to follow, examples." %}</em>
+              <em>{% trans "(Optional) Define AI model's role, background information, and rules to follow." %}</em>
```
smcdonald-jus, line 133: Removed "examples" because I've always included those in pre-query instructions, not system prompt. Is that just me?

django/chat/templates/chat/modals/advanced_qa_modal.html
```diff
@@ -160,7 +160,7 @@ <h5 class="modal-title" id="advanced-qa-modal-title">
           <div class="col-lg-4 mb-2">
             <label for="qa_post_instructions-modal" class="col-form-label-sm pb-0">{% trans "Post-query instructions" %}</label>
             <p class="text-secondary small mb-2">
-              <em>{% trans "(Optional) Final reminders of desired behaviour. Keep this short." %}</em>
+              <em>{% trans "(Optional) Final reminders of desired behaviour. Keep this fairly short." %}</em>
```
smcdonald-jus, line 163: Added qualifier because the default post-query instructions are, imo, not "short" but "_fairly_ short"

django/chat/templates/chat/modals/advanced_qa_modal.html
juleskuehn: Good point. We actually have examples in the default pre-query instructions - so that makes sense!

PR #285, opened 2025-03-18
feat: zip-loading

amusbag:
Uploading zip files and extracting the contents and recursively uploading them.


django/librarian/utils/process_engine.py
```diff
@@ -120,6 +122,7 @@ def guess_content_type(
         "application/pdf",
         "application/xml",
         "application/vnd.ms-outlook",
+        "application/x-zip-compressed",
```
juleskuehn, line 125: Why not `application/zip`? I notice that if I upload a ZIP that contains other ZIPs, `application/zip` is the one that is detected. (And these do extract, by the way).

django/librarian/utils/process_engine.py
```diff
@@ -742,3 +753,35 @@ def resize_to_azure_requirements(content):
             image.save(output, format="PNG")
             content = output.getvalue()
             return content
+
+
+def process_file(file, data_source_id, name, content_type):
+    from librarian.utils.process_engine import generate_hash
+
+    file_hash = generate_hash(file.read())
+    file_exists = SavedFile.objects.filter(sha256_hash=file_hash).exists()
```
juleskuehn, line 762: This doesn't seem to be working. There are a bunch of redundant files getting created.

amusbag: I will look into that right now.

django/librarian/utils/extract_zip.py
```diff
@@ -0,0 +1,35 @@
...
+                    content_type = guess_content_type(f, name)
+                    process_file(f, data_source_id, name, content_type)
+                file_info = f"Filename: {info.filename}\nModified: {datetime.datetime(*info.date_time)}\nNormal size: {info.file_size} bytes\nCompressed size: {info.compress_size} bytes\n--------------------\n"
+                info_list.append(file_info)
+            md = f"".join(info_list)
+        except Exception as e:
+            logger.error(f"Failed to extract Zip file: {e}")
+            md = ""
+        shutil.rmtree(directory, ignore_errors=True)
+        return md
```
juleskuehn, line 35: I'd like to see the Document objects for the ZIP files deleted after they are finished extracting.

django/librarian/utils/process_engine.py
amusbag: It should be application/zip, the initial one tends to come in as application/x-zip-compressed so I was testing something else and forgot to change it back to application/zip. I will make that change when I make the fixes.

juleskuehn (commented):
We had discussed including a longer "file path" property which would keep track of the hierarchy. e.g. if you have a ZIP file within a zip file, then a pdf within that, the file path should be:
`documents.zip/nested.zip/lorem_ipsum.pdf`

This would be reused for similar "nested extraction" functionality like email attachments or folder uploads. I expect you would need to pass around a "path_prefix" attribute when you make the recursive calls to "process_file".

django/librarian/utils/process_engine.py
juleskuehn: It can be both - doesn't seem to make a difference - just wanted to make sure you had considered this. Maybe both should be explicitly included.

juleskuehn (approved):

PR #286, opened 2025-03-20
feat: address split

Rifahaziz:


juleskuehn (approved):
Didn't test it locally, but this seems reasonable :)

If you are running an LLM call on the full document for each question though, I think you can optimize that quite a bit (this is a costly way to extract a street number from a document). I would use structured outputs to ensure it returns certain types of data. We can do this in a follow-up PR though.

Rifahaziz:
Sure, I can do that once it is sure this project is ours. Or I can keep a PR open for it now too. thanks

PR #287, opened 2025-03-20
fix: sync_users recurring task

haitmo:
Addresses failing task in k8s.

- [x] Celery beat tested locally
- [x] Celery beat tested in sandbox


django/otto/celery.py
```diff
@@ -49,12 +49,12 @@
     },
     # Delete unused libraries every day at 3 am UTC
     "delete-unused-libraries-every-morning": {
-        "task": "otto.task.delete_unused_libraries",
+        "task": "otto.tasks.delete_unused_libraries",
```
haitmo, line 52: Unrelated to PR.

But I don't think these were working considering the typo.

juleskuehn (approved):

PR #288, opened 2025-03-20
feat: add docx page numbers

Rifahaziz:
**Not to merge yet**

Detecting page break is working for simple word documents. Inaccurate page numbers for the following cases:
-When there are opening pages and page 1 starts after ToC - it will detect the first page as page 1 whereas in reality page 1 starts after ToC (can be after 5-6 pages)
-Page breaks are not accurate for empty pages or pages with layouts - it merged three title pages as one page.
-Page break detection is incorrect when there are tables - it merged two pages into one.

For a word document with 56 pages that has layouts and tables, it detected total pages as 39 due to merging certain pages.
Best way to deal with formatting/layout with correct page numbers is to convert it first to a pdf. So, should we keep what we have for simple word documents and apply pdf conversion for documents with layouts?

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -80,26 +80,4 @@
   ],
   "python.testing.unittestEnabled": false,
   "python.testing.pytestEnabled": true,
-  "workbench.colorCustomizations": {
-    "terminal.background": "#2E3440",
-    "terminal.foreground": "#E5E9F0",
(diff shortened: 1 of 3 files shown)


juleskuehn:
@Rifahaziz if you agree, can you close this PR due to lack of accuracy with page numbers? Thanks.

PR #293, opened 2025-03-26
fix: session timeout (ignore user_cost polling)

juleskuehn:
To test:

In settings.py, set `SESSION_COOKIE_AGE = 305`.
In user_cost.html, set `hx-trigger='load delay:2000ms'

The warning will appear within a few seconds, if you don't do anything.

### DIFF
--- a/django/otto/settings.py
+++ b/django/otto/settings.py
@@ -176,10 +176,10 @@ MIDDLEWARE = [
     "whitenoise.middleware.WhiteNoiseMiddleware",
     "django_prometheus.middleware.PrometheusBeforeMiddleware",
     # SC-10, SC-23: Django default session management
-    "otto.utils.middleware.ExtendSessionMiddleware",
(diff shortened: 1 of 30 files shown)


django/otto/settings.py
```diff
@@ -93,10 +93,6 @@
 LOGIN_URL = "/azure_auth/login"
 LOGIN_REDIRECT_URL = "/"  # Or any other endpoint
 
-# Session timeout. 24 hours is allowed for WCAG and meets security requirement
-SESSION_COOKIE_AGE = 60 * 60 * 24  # 24 hours, in seconds
-SESSION_SAVE_EVERY_REQUEST = True  # Reset the timeout on every request
```
juleskuehn, line 98: These were duplicated later in the file.

haitmo (approved):

haitmo (approved):

PR #296, opened 2025-03-27
feat: prevent concurrent logins

juleskuehn:
To resolve VA feedback 2.14

To test, open two different browsers and alternate between them trying to use Otto.


haitmo (approved):
Awesome thanks for doing this!

PR #299, opened 2025-03-27
chore: Lex experiment uses single prompt now with pydantic class

Rifahaziz:


juleskuehn (commented):
Great - this is an improvement in quality and speed over the previous version! However, in both the previous version AND in this version, you are still using the OpenAI API directly rather than going through the OttoLLM class. So costs are not being recorded except for the OCR. The way you are doing it (just create a cost object) is not correctly tracking the number of tokens used. That is what the OttoLLM class takes care of.

One issue with the performance is the extraction of section references. Sometimes the Notices of Appeal reference both Income Tax Act and Income Tax Regulations. It is not clear in the new version which is which. You might need to break this into 2 columns. The formatting consistency is greatly improved though.

I haven't noticed any parsing errors with the files you provided, but it is possible you will have them with other files. Specifically for the filing date which isn't actually validated in date format.

You shouldn't have to write out the system prompt yourself. You can use features in LlamaIndex to do this (which will also leverage JSON mode in OpenAI to ensure it follows the format).
https://docs.llamaindex.ai/en/stable/understanding/extraction/structured_llms/
This will take care of the cost tracking (provided you access the llm object through OttoLLM() class) as well as not needing to repeat your schema in both Pydantic and JSON formats.

Rifahaziz:
Putting this off for a while until the important bugs are fixed.

juleskuehn:
Since this is an improvement over the last version, I am going to merge it. Can address my comments later if necessary.

PR #303, opened 2025-03-28
fix: always use Azure OCR for PDFs (no PDFium)

juleskuehn:
PDFium is not thread safe which created issues during the stress test.
For now, we will just not use it at all.


amusbag (approved):
Looks good! Tests passed locally.

PR #311, opened 2025-04-03
fix: librarian bug

juleskuehn:
Bug introduced in #302 
- creating/updating any data source except chat uploads would result in 500 error

Another way to solve the issue, which I pursued here:
- don't allow users to modify the chat upload folders from within librarian

A couple related improvements:
- only show chats that have Q&A uploads in librarian
- only list chats that have Q&A uploads in the "Select chat" autocomplete widget (in Q&A settings sidebar)
  - plus "This chat" always shown too, regardless of whether it has any Q&A uploads yet.


Rifahaziz (approved):
Thank you Jules!

PR #314, opened 2025-04-07
fix: compile translation files during Docker build

haitmo:
Localization do not occur in DEV. I've included them in the build instead as they shouldn't really change anyways.


haitmo:
Will handle this in the infrastructure repo

PR #317, opened 2025-04-07
fix: pypdfium lock

juleskuehn:
Also reverts the removal of Pdfium as default mode for Q&A loading.

Needs to be tested in sandbox with multiple users to ensure no Pdfium errors.


amusbag (approved):
Looks good. Runs locally and tests passed.

PR #321, opened 2025-04-07
fix: restore chat title and security label updating

juleskuehn:
Fixes a regression introduced in #301


django/chat/templates/chat/components/chat_security_label.html
```diff
@@ -1,7 +1,9 @@
 
 {% load i18n %}
 
-<div class="dropdown">
+<div class="dropdown"
+     id="chat-security-label-{{ chat.id }}"
+     {% if swap_oob %}hx-swap-oob="true"{% endif %}>
```
juleskuehn, line 6: The ID and hx-swap were needed for security label to update when a Folder with security > chat security is queried in RAG mode (it's not a very important feature, but might as well fix it).

django/chat/utils.py
```diff
@@ -293,6 +293,7 @@ def sse_string(
             title_llm = OttoLLM()
             await sync_to_async(title_chat)(chat.id, force_title=False, llm=title_llm)
             await sync_to_async(title_llm.create_costs)()
+            await sync_to_async(message.chat.refresh_from_db)()
```
juleskuehn, line 296: This solved a problem with chat titles not updating. (They should update in the sidebar after a message or two)

Rifahaziz (approved):

PR #323, opened 2025-04-08
fix: check for none messages/responses

juleskuehn:
Fixes a bug in main :| which seems to have been introduced by llama_index upgrade...


amusbag (approved):
Looks good. Tested locally and all systems go.

PR #324, opened 2025-04-08
fix: htmx redirect checks

juleskuehn:
Previously some middleware/decorator checks would return a redirected view inside a HTMX request. This resulted in a "Otto within Otto" swap inside e.g. a modal that was making a not-permitted HTMX request.


haitmo (approved):
Good catch!

PR #328, opened 2025-04-08
fix: tldextract file path

juleskuehn:
To test:
* Delete your local django/tld_cache directory
* Restart django
* Add a URL to Q&A mode (bad or good url)
* Ensure tld_cache directory is created with no errors in the django logs


amusbag (approved):
Looks good. tld_cache created with no errors in console.

PR #334, opened 2025-04-14
fix: code block dark mode

amusbag:
Added chat-message-text-color variable
Changed dark mode color of chat-message-bg


juleskuehn (approved):
Looks great! Thanks.

PR #335, opened 2025-04-15
style: Increased contrast of search bar

andrewkong9:


juleskuehn:
Thanks Andrew. Make sure not to apply this style in dark mode. I don't really like the look of it in light mode either, but it does accomplish the goal of the ticket!

django/otto/static/css/variables.css
```diff
@@ -149,6 +150,7 @@
   --search-bar-hover-bg: #3e474f;
   --chat-delete-text: #ff5e5e;
   --prompt-send-btn-bg: #1a1e20;
+  --search-bar-border: #808080;
```
juleskuehn, line 153: It would make more sense to just change this to `var(--bs-border-color);` and then leave style.css as it was before.

andrewkong9: I am a little ensure how to do this since if I were to change the border-color to var(--search-bar-border), it would always yield to the dark mode colour since the top bar is permanently in dark mode. I am unsure how to make the change to recognize the first theme, rather than the second when determining a color. 

In light mode, it would be [data-bs-theme="light"] [data-bs-theme="dark"]
In dark mode, it would be [data-bs-theme="dark"] [data-bs-theme="dark"]

The colours are only determined by the second instance of data-bs-theme

juleskuehn (dismissed):

juleskuehn (approved):

PR #337, opened 2025-04-16
fix: summarize bad file

amusbag:
Add corrupt files array to keep track of the corrupt files
Wrap the function in a try block
Remove corrupt file name from titles in the except block and append it to the corrupt files array
Log the error
Continue to the next file
Create corrupt file string with all the corrupt files and append back to titles to present to user with the error message

### DIFF
--- a/django/chat/responses.py
+++ b/django/chat/responses.py
@@ -173,9 +173,22 @@ def summarize_response(chat, response_message):
     if len(files) > 0:
         titles = [file.filename for file in files]
         responses = []
+        corrupt_files = []
         for file in files:
             if not file.text:
-                file.extract_text(pdf_method="default")
+                try:
+                    file.extract_text(pdf_method="default")
+                except Exception as e:
+                    titles.remove(file.filename)
+                    corrupt_files.append(file.filename)
+                    logger.error(
+                        "Error extracting text from file",
+                        error_id=str(uuid.uuid4())[:7],
+                        message_id=response_message.id,
+                        chat_id=chat.id,
+                        error=traceback.format_exc(),
+                    )
+                    continue
             if not cache.get(f"stop_response_{response_message.id}", False):
                 responses.append(
                     summarize_long_text(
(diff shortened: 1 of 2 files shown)


amusbag:
I wanted you to review how the error is logged, I used the copilot suggestion.

django/chat/responses.py
```diff
@@ -169,26 +170,43 @@ def summarize_response(chat, response_message):
...
 
     llm = OttoLLM(model)
+    error_str = ""
 
     if len(files) > 0:
         titles = [file.filename for file in files]
         responses = []
         for file in files:
+            if cache.get(f"stop_response_{response_message.id}", False):
+                break
```
juleskuehn, line 180: Moving this up here will break faster; reversing the condition so we can lose a level of indentation

django/chat/responses.py
```diff
@@ -169,26 +170,43 @@ def summarize_response(chat, response_message):
...
-                        gender_neutral,
-                        instructions,
+                try:
+                    file.extract_text(pdf_method="default")
+                except Exception as e:
+                    error_id = str(uuid.uuid4())[:7]
+                    error_str = _(
+                        "Error extracting text from file. Try copying and pasting the text."
                     )
+                    error_str += f" _({_('Error ID')}: {error_id})_"
```
juleskuehn, line 189: Showing error ID to users to help us with debugging

django/chat/responses.py
```diff
@@ -169,26 +170,43 @@ def summarize_response(chat, response_message):
...
+                    responses.append(stream_to_replacer([error_str]))
```
juleskuehn, line 190: Instead of having a separate list of errors, output them under the existing filename headings

django/chat/responses.py
```diff
@@ -211,44 +229,47 @@ def summarize_response(chat, response_message):
...
     else:
+        # Text input is a URL or plain text
         url_validator = URLValidator()
         try:
             url_validator(user_message.text)
             text_to_summarize = url_to_text(user_message.text)
+            # Check if response text is too short (most likely a website blocking Otto)
+            if len(text_to_summarize.split()) < 35:
+                error_str = _(
+                    "Couldn't retrieve the webpage. The site might block bots. Try copy & pasting the webpage here."
```
juleskuehn, line 242: This logic was poorly nested before so it would sometimes occur even when user hadn't entered a URL.

django/chat/responses.py
```diff
@@ -211,44 +229,47 @@ def summarize_response(chat, response_message):
...
+                )
         except ValidationError:
             text_to_summarize = user_message.text
 
-        # Check if response text is too short (most likely a website blocking Otto)
-        if len(text_to_summarize.split()) < 35:
-            summary = _(
-                "Couldn't retrieve the webpage. The site might block bots. Try copy & pasting the webpage here."
-            )
-        else:
```
juleskuehn, line 228: The rest is pure refactoring; doesn't change behaviour

PR #338, opened 2025-04-17
chore: error messages 

Rifahaziz:
1. More try-catch blocks included that has:
   - error logs for developers
   - simple error messages to users
2. Code is refactored accordingly, tests are adjusted. I tried to separate when the error is from the user's side vs our side. 

Progress of applying 1 and 2 to:
- [x] Text Extractor
- [x] AI Assistant
- [x] Legislation Search
- [x] Otto general


django/chat/views.py
```diff
@@ -289,6 +291,14 @@ def chat_message(request, chat_id):
         entered_url = True
         allowed_url = check_url_allowed(user_message_text)
     except ValidationError:
+        full_error = traceback.format_exc()
+        error_id = str(uuid.uuid4())[:7]
+        logger.error(
+            f"Validation error in user message:",
```
juleskuehn, line 297: This one should not log an error - it's just checking for a URL. Just get rid of the changes in this block please.

django/chat/utils.py
```diff
@@ -313,7 +319,9 @@ def sse_string(
 
     except Exception as e:
         message = await sync_to_async(Message.objects.get)(id=message_id)
-        full_message = _("An error occurred.")
+        full_message = _(
+            "An error occurred."
+        )  # should this be changed to logging error?
```
juleskuehn, line 324: Yes, this should be made to be similar to error_response():
```
def error_response(chat, response_message, error_message=None):
    """
    Send an error message to the user.
    """
    llm = OttoLLM()
    response_str = _("There was an error processing your request.")
    error_id = str(uuid.uuid4())[:7]

    if error_message and settings.DEBUG:
        response_str += f"\n\n```\n{error_message}\n```\n\n"
    response_str += f" _({_('Error ID')}: {error_id})_"
    logger.error(
        "Error processing chat response",
        error_id=error_id,
        message_id=response_message.id,
        chat_id=chat.id,
        error=traceback.format_exc(),
    )
    return StreamingHttpResponse(
        streaming_content=htmx_stream(
            chat,
            response_message.id,
            llm,
            response_str=response_str,
        ),
    )
```

We should possibly remove the check for "settings.DEBUG" - I'm checking with Hamza.

django/otto/views.py
```diff
@@ -452,6 +455,10 @@ def manage_users_upload(request):
                         monthly_max = int(monthly_max)
                     except Exception as e:
                         monthly_max = None
+                        logger.info(
+                            f"monthly_max is not an integer ({monthly_max}): {e}"
```
juleskuehn, line 459: This one doesn't require logging .

django/text_extractor/tasks.py
```diff
@@ -22,14 +25,41 @@ def process_ocr_document(file_content, file_name, merged, idx):
...
+        full_error = traceback.format_exc()
+        error_id = str(uuid.uuid4())[:7]
+        logger.error(
+            f"Error processing file {file_name} in task {current_task.request.id}: {full_error}"
+        )
+        # Fallback for other exceptions
+        return {
+            "error": True,
+            "full_error": full_error,
+            "message": f"Error: Check your file for corruption of type mismatch",
```
juleskuehn, line 63: This should be translated if kept. But I think since this ends up in a table view, you don't actually want a long text. No need to change the way errors are displayed in Text Extractor IMO, except perhaps to show the user the Error ID (to help us in debugging).

django/text_extractor/utils.py
```diff
@@ -209,105 +266,121 @@ def create_searchable_pdf(input_file, add_header, merged=False):
...
+            # Create a new PDF page
+            new_pdf_page = PdfReader(ocr_overlay)  # changed
+            output.add_page(new_pdf_page.pages[0])
+    except Exception as e:
+        full_error = traceback.format_exc()
+        error_id = str(uuid.uuid4())[:7]
+        logger.error(f"Error creating PDF overlay after OCR {error_id}: {full_error}")
+        return {
+            "error": True,
+            "message": f"Error: Failed to create PDF overlay after OCR.",
```
juleskuehn, line 372: Same for all messages - ensure they are translated. Not sure if I caught them all in the review but please check.

juleskuehn (commented):
Thanks for adding the logging, Rifah. 
We only need error_id generated and logged when the error_id is displayed to users.

If there is no way to display the error ID to the user, no need to generate or log the Error ID. If there *is* a way to display to error ID, please make sure you do.

See chat/responses/error_response() which has the desired implementation (except we actually do want to show it in all cases, not just when settings.DEBUG - I will get rid of that condition)

django/chat/utils.py
Rifahaziz: done.

django/text_extractor/tasks.py
Rifahaziz: error id shown and message shortened.

django/otto/views.py
Rifahaziz: removed

django/text_extractor/utils.py
Rifahaziz: fixed all translations

django/chat/responses.py
```diff
@@ -176,9 +175,9 @@ def chat_response(
                     "2. Using summarize mode, which can handle longer texts\n"
                     "3. Using a different model\n"
                 ),
+                switch_mode=switch_mode,
```
juleskuehn, line 178: Fixes rare bug

django/chat/responses.py
```diff
@@ -464,8 +472,10 @@ async def add_files_to_library():
         )()
         if error_documents:
             error_string = _("Error processing the following document(s):")
-            doc_names_for_error = [doc.filename for doc in error_documents]
-            error_docs_joined = "\n\n - " + "\n\n - ".join(doc_names_for_error)
+            doc_errors = [
+                f"{doc.filename} _{doc.status_details}_" for doc in error_documents
```
juleskuehn, line 476: This shows the Error IDs for documents that didn't process correctly when files are uploaded in the chat to Q&A mode.

juleskuehn (approved):
Thanks, this looks good now and I've tried to manually test all the areas I can.

PR #340, opened 2025-04-23
fix: markdown <page_x> tags not closing

amusbag:
Fix error where the page tags for uploads not always closing properly.

### DIFF
--- a/django/librarian/utils/markdown_splitter.py
+++ b/django/librarian/utils/markdown_splitter.py
@@ -102,7 +102,10 @@ class MarkdownSplitter:
         html_string = html_string.strip()
         # Closing tag as the first line?
         if html_string.split("\n", 1)[0].startswith("</page_"):
-            html_string = html_string.split("\n", 1)[1]
+            if 1 in range(len(html_string.split("\n", 1))):
+                html_string = html_string.split("\n", 1)[1]
+            else:
+                html_string = html_string = html_string.split("\n", 1)[0]
         # Opening tag as the last line?
         if html_string.rsplit("\n", 1)[-1].startswith("<page_"):
             html_string = html_string.rsplit("\n", 1)[0]
--- a/django/librarian/utils/markdown_splitter.py
+++ b/django/librarian/utils/markdown_splitter.py
@@ -100,12 +100,14 @@ class MarkdownSplitter:
         * chunk ends with <page_n+1> (with no content for that page)
         """
         html_string = html_string.strip()
+        print(f"HTML string: {html_string}")
         # Closing tag as the first line?
         if html_string.split("\n", 1)[0].startswith("</page_"):
             if 1 in range(len(html_string.split("\n", 1))):
                 html_string = html_string.split("\n", 1)[1]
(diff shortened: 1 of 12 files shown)


django/librarian/utils/markdown_splitter.py
```diff
@@ -100,9 +100,14 @@ def _close_page_tags(self, html_string: str) -> str:
         * chunk ends with <page_n+1> (with no content for that page)
         """
         html_string = html_string.strip()
+        print(f"HTML string: {html_string}")
         # Closing tag as the first line?
         if html_string.split("\n", 1)[0].startswith("</page_"):
-            html_string = html_string.split("\n", 1)[1]
+            if 1 in range(len(html_string.split("\n", 1))):
```
juleskuehn, line 106: The condition `if len(html_string.split("\n", 1) > 1` may be more clear.

This is the default case right? Where a closing page tag starts the chunk, but there is more content below, so the initial closing tag is simply removed?
Might help to add some comments. (I should have done so before).

django/librarian/utils/markdown_splitter.py
```diff
@@ -100,9 +100,14 @@ def _close_page_tags(self, html_string: str) -> str:
...
+                html_string = html_string.split("\n", 1)[1]
+            else:
+                html_string = html_string.split("\n", 1)[0]
+        # Opening tag as the first line?
```
juleskuehn, line 110: Why is this comment here?

django/librarian/utils/markdown_splitter.py
```diff
@@ -100,9 +100,14 @@ def _close_page_tags(self, html_string: str) -> str:
...
         """
         html_string = html_string.strip()
+        print(f"HTML string: {html_string}")
         # Closing tag as the first line?
         if html_string.split("\n", 1)[0].startswith("</page_"):
-            html_string = html_string.split("\n", 1)[1]
+            if 1 in range(len(html_string.split("\n", 1))):
+                html_string = html_string.split("\n", 1)[1]
+            else:
+                html_string = html_string.split("\n", 1)[0]
```
juleskuehn, line 109: This is the case where the chunk ONLY contains a closing tag? In this case, shouldn't the chunk just be removed?

django/librarian/utils/markdown_splitter.py
```diff
@@ -100,9 +100,14 @@ def _close_page_tags(self, html_string: str) -> str:
         * chunk ends with <page_n+1> (with no content for that page)
         """
         html_string = html_string.strip()
+        print(f"HTML string: {html_string}")
```
juleskuehn, line 103: Remove print statement when finished.

juleskuehn (commented):
Thanks for the fix, seems like this will work! Thanks for adding the test.

django/librarian/utils/markdown_splitter.py
amusbag: Copilot probably added that.

juleskuehn (approved):

PR #341, opened 2025-04-23
fix: librarian bugs, fr-ca translation

juleskuehn:


django/chat/responses.py
```diff
@@ -400,15 +401,14 @@ async def add_files_to_library():
                 lambda: ds.documents.filter(status__in=["INIT", "PROCESSING"]).count()
             )()
 
-        saved_files = [file.saved_file for file in files]
         error_documents = await sync_to_async(
             lambda: list(
-                ds.documents.filter(status="ERROR", saved_file__in=saved_files)
+                ds.documents.filter(status="ERROR", created_at__gt=message_created_at)
```
smcdonald-jus, line 406: What's the rationale behind replacing the saved_file filter with one based on creation times? Intuitively it seems like the former would be more appropriate.

juleskuehn: zip files will create more documents that otherwise wouldn't be counted because they weren't directly uploaded to the chat message.
also, zip files are deleted after processing so the count would otherwise say "0" for a zip file.

smcdonald-jus (approved):

PR #342, opened 2025-04-25
feat: warning for high cost requests

mcusteau:


juleskuehn:
Because I changed the default cost threshold, the test now fails. Let's set WARN_COST as a constant in settings.py and then reference that in both the function and the tests.

django/chat/utils.py
```diff
@@ -959,3 +979,72 @@ def get_chat_history_sections(user_chats):
...
+                    for doc in data_source.documents.all():
+                        cost += estimate_cost_of_string(
+                            doc.extracted_text, model + "-in"
+                        )
+    elif mode == "summarize" or "translate":
+        model = chat.options.summarize_model
+        for file in response_message.parent.sorted_files.all():
+            if not file.text:
+                try:
+                    file.extract_text(pdf_method="default")
```
smcdonald-jus, line 1035: Is there any associated cost with running this here? The cost estimation shouldn't incur costs itself, imo.

django/chat/responses.py
```diff
@@ -84,16 +97,32 @@ def otto_response(request, message_id=None, switch_mode=False, skip_agent=False)
...
+            "chat/components/cost_warning_buttons.html",
+            {
+                "message_id": response_message.id,
+                "estimate_cost": formatted_cost,
+                "continue_button": True,
+            },
+        ).replace("\n", "")
 
     llm = OttoLLM()
     # This will only be reached in an error case
```
smcdonald-jus, line 120: I think this comment is obsolete, no?

juleskuehn (approved):
Working great!

juleskuehn (approved):

PR #344, opened 2025-04-28
fix: js chunked uploads on k8s replicas

juleskuehn:
Currently, uploading files to a chat can result in errors, likely due to concurrent or out-of-sequence writes from multiple pods handling different chunks of the same file upload.

In this PR, we write each chunk to a separate file, then on the final request to "done_upload", reassemble them into the single file associated with the ChatFile.file property.

Still needs a bit of cleanup.


amusbag (approved):
Looks good.

PR #362, opened 2025-05-03
fix: JS chat uploads (use django-file-form)

juleskuehn:
* needs tests
* issue with dangling SavedFile / redundant files persists
  * this is related to zip extraction functionality and not related to the JS upload issue
  * for now, i've just added a daily cleanup task


amusbag (approved):
Looks good.

PR #365, opened 2025-05-05
fix: french translations; french sharing icon not displaying properly

mcusteau:
fix some french translations and french sharing icon not displaying properly


haitmo (approved):
Thanks Michel! Looks good

PR #376, opened 2025-05-06
fix: replace native form submission with htmx for file uploads in Firefox

haitmo:


juleskuehn (approved):
Working great in Edge as well.

PR #380, opened 2025-05-07
fix: options sidebar / preset loading bugs

juleskuehn:
When loading a preset, it appeared to work, but the old chat options were being saved again right after loading - thus making the loaded preset NOT actually function.


Rifahaziz (commented):
Few issues:
-If I click "Edit libraries" button while on my loaded preset, it resets 'search mode' and 'Filter results'
-summary desire language not taking into account in the backend, for both when preset is loaded AND when you manually change the language.
-asking for a table in summary mode messes up the response (but we can ignore this for now)

smcdonald-jus (approved):

PR #384, opened 2025-05-12
feat: eml-loading

amusbag:
- added function to extract content of eml files

This was tested with multiple eml files including files with attachments. Not tested with eml file as attachment within another eml file as Outlook (classic) doesn't allow emails to be saved as eml so I couldn't create one. Creating as draft until that is sorted.


juleskuehn (approved):
Looks good Aliyu! Thanks.

PR #388, opened 2025-05-12
empty string replaced with None

mcusteau:


django/chat/views.py
```diff
@@ -723,7 +723,7 @@ def replace_page_tags(match):
     ).filter(message_id=message_id):
         source_text = str(source.node_text)
 
-        already_processed = source.processed_text != ""
+        already_processed = source.processed_text != None
```
juleskuehn, line 726: I'm sure this works but you should technically check equality with the "None" object using "is". i.e. `is not None`

juleskuehn (approved):
thanks!

PR #390, opened 2025-05-13
fix: errors and successes counted properly from .zip files

smcdonald-jus:
The old logic to display the "files ready for Q&A" message broke in the presence of _exactly one_ corrupted file in a .zip. This is because the logic relied on comparing the number of error docs to the number of _files_, and a .zip counts as only one file despite resulting in multiple docs. I've made the logic more direct, but can't say with confidence whether there's a good reason _not_ to do it this way.


juleskuehn:
Looks good!

juleskuehn (approved):

PR #391, opened 2025-05-13
fix: 0 documents added

mcusteau:
current code doesn't add duplicate documents to total count of uploaded docs, which is technically accurate but confusing to user


juleskuehn (approved):
Looks good and working properly with all cases I tested. Thanks!

PR #419, opened 2025-06-24
chore: update .env.example; robustness fixes

juleskuehn:
Addresses some issues I encountered when deploying Sandbox, but applies to the main branch as well.
* update endpoints in .env.example
* handle models that only work with a single prompt
* fill empty messages in chat history (e.g. with filenames uploaded by user or created by bot)
* make chat/utils.py "copy_options" robust to use from within management commands (for loading preset fixtures)
* update "reset_app_data --all" management command to load the library_mini.yaml


amusbag (approved):
Looks good.

PR #433, opened 2025-07-09
fix: feedback dashboard changes

Rifahaziz:
1. Pagination issue fixed: Originally there was an issue where the filters were not being applied to any pages except the first page. Now this is fixed by using GET instead of POST request in the filter form.
2. A new option of "deferred" added in the filter.
3. More stats added in the top header bar: Counts for New, In progress, Resolved, Closed and Deferred


django/otto/templates/components/feedback/dashboard/feedback_stats.html
```diff
@@ -1,36 +1,68 @@
...
+      <div class="col-6 col-md-3 col-lg-2 p-2 border-end border-bottom">
+        <div class="d-flex justify-content-between align-items-center">
+          <p class="text-muted fw-semibold mb-1">{% trans "Total closed" %}</p>
+          <!-- Add stats later if necessary -->
+          <!-- <span class="fw-semibold">+12%</span>-->
+        </div>
+        <p style="letter-spacing: -0.025em;" class="display-6 fw-semibold mb-0">{{ stats.closed }}</p>
+      </div>
+
+      <div class="col-6 col-md-3 col-lg-2 p-2 border-end border-bottom">
```
Rifahaziz, line 56: just making it look aligned with a scroll bar

django/otto/views.py
```diff
@@ -240,20 +241,23 @@ def feedback_list(request, page_number=None):
...
-
-        if feedback_type and feedback_type != "all":
-            feedback_messages = feedback_messages.filter(feedback_type=feedback_type)
-        if status and status != "all":
-            feedback_messages = feedback_messages.filter(status=status)
-        if app and app != "all":
-            feedback_messages = feedback_messages.filter(app=app)
-
-    # Get 10 feedback messages per page
+    # Only use GET for filters to ensure pagination works with query params
```
Rifahaziz, line 244: "POST" changed to "GET" here mainly

Rifahaziz:
Will try 1. For 2 - I thought closed means no work needed to be done (i.e maybe its just a comment or idea) while resolved meant work has been done and issue is solved. That's what came to my mind. If it makes sense to merge them into one, sure. Let me know what you think.

smcdonald-jus:
Your distinction between "Closed" and "Resolved" makes sense to me! Let's keep them separate then. I would still change the left-to-right order, though.

smcdonald-jus (approved):

PR #439, opened 2025-07-15
fix: laws schedules details page not working

amusbag:
- issue with details page not working due to some characters in node_id being replaced with entirely different ones.
 - node_id: urllib.parse.quote_plus(s.node.node_id).replace("+", "-") was causing the issue
 - solution was to check if node_id contains "_schedule_" and not make any replacements if it does


haitmo (approved):
Seems to work great! Thanks

PR #440, opened 2025-07-16
chore: update deps

juleskuehn:
Doesn't solve or cause issues here; on sandbox branch this fixes an issue though, and we'll have to update eventually..


juleskuehn:
Llama Index new version has some issues. Waiting for them to be resolved before upgrading.

PR #441, opened 2025-07-21
feat: add per-doc RAG mode, refactor organization of Q&A options

smcdonald-jus:
There is now a mode that runs RAG on each selected document separately.
While implementing this, it seemed that it would be more sensible to reorganize the Q&A options. Instead of the previous 3 modes (RAG, full doc/separate, full doc/combined), there are now two binary choices the user can make:

1. "Top excerpts" (RAG) or "Full documents" (tree summarizer)
2. "Combined" (consider all selected documents at once when answering) or "Separate" (per-doc answers)

Note that granularity is enabled for both RAG options. When running "combined RAG", this groups sources _after_ retrieving them, with whichever document groupings happen to be possible with the most relevant retrieved sources. With "separate RAG", the sources are retrieved separately for _each_ selected document, and then grouped according to the granularity logic. This case results in roughly `num_documents * num_granular_groups_per_doc` answers.

Various other formatting changes, refactorings, and model cleanup happened to enable this.


juleskuehn (dismissed):
Looks great Shaun. I think this is a big improvement!

Some of the wording could be improved:
Answer mode (not great): Documents together (even less clear?)
Search filters: Well, it isn't plural. How about Filter documents? Not sure. Please brainstorm. Otto will help :) I would move this field right under "Library" select.

Other than this and the minor code quibble, I think this is a step up in clarity and ready to go.

django/chat/responses.py
```diff
@@ -546,199 +546,92 @@ async def add_files_to_library():
...
+                    chat, response_message, llm, [document], qa_scope, switch_mode
+                )
 
-            doc_groups = group_sources_into_docs(source_nodes)
+                if answer_components:
+                    (
+                        doc_response_replacer,
+                        doc_response_generator,
+                        doc_source_groups,
+                        doc_batch_generators,
```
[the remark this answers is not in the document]
smcdonald-jus: Good call!

django/chat/forms.py
```diff
@@ -303,7 +304,28 @@ class Meta:
                 choices=QA_MODE_CHOICES,
                 attrs={
                     "class": "form-select form-select-sm",
-                    "onchange": "switchToDocumentScope(); updateQaSourceForms(); toggleRagOptions(this.value); triggerOptionSave();",
+                    "onchange": "switchToDocumentScope(); updateQaSourceForms(); toggleRagOptions(this); triggerOptionSave();",
+                    "data-rag_string": _(
+                        """
+                        Combine: search once across all selected documents before answer generation. <em>May not include all documents. Cheap, more succint.</em>
```
smcdonald-jus, line 310: Note that I went with "search once _before answer generation_" instead of "search once _and write a single answer_". You could put these settings on and then enable granularity, in which case "single answer" isn't true. "Before answer generation" is general enough to apply regardless of granularity.

django/chat/migrations/0016_remove_chatoptions_qa_answer_mode_and_more.py
```diff
@@ -0,0 +1,69 @@
...
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("chat", "0015_alter_chatoptions_chat_reasoning_effort"),
+    ]
+
+    def migrate_chatoptions_data(apps, schema_editor):
```
smcdonald-jus, line 12: Took the easy way out and let GPT 4.1 whip up this data migration so that user presets are preserved with the new model structure

django/chat/responses.py
```diff
@@ -546,199 +546,92 @@ async def add_files_to_library():
...
+                        else doc_response_generator
+                    )
 
-            if chat.options.qa_source_order == "reading_order":
-                # Reading order requires keeping docs together, so
-                # sort documents by maximum node score within doc
-                # before stitching nodes together
-                doc_groups = sort_by_max_score(doc_groups)
+                    source_groups.extend(doc_source_groups)
+                    batch_generators.extend(doc_batch_generators)
```
smcdonald-jus, line 593: Note: because of the way batching/generators/replacers are all set up, streaming happens _one document at a time_ with granular per-doc RAG. Fixing that would require some kind of nested batching. That'd be quite complicated due to the way I previously set up parallel streaming and delineation between batches.

Thus, in this case, the batching happens only for the granular groups within each doc, and the doc streams are just glued together. We can explore this in a future ticket, but granular per-doc RAG is niche enough that optimized streaming is very low priority, imo.

Note that parallel batch streaming _does_ happen with per-doc RAG when granularity is _off_.

django/chat/responses.py
```diff
@@ -755,6 +648,194 @@ async def add_files_to_library():
...
+
+        # Stitching
+        for doc in doc_groups:
+            current_source_group = []
+            for next_source in doc:
+                if num_tokens_from_string(
+                    "\n\n".join(
+                        [x.text for x in current_source_group] + [next_source.text],
+                    ),
+                    "cl100k_base",
```
smcdonald-jus, line 770: Previously, this function counted the tokens with the o200k_base encoding. However, I'm pretty sure that uploaded files are chunked with cl100k_base. If so, this change makes it more consistent.

django/chat/templates/chat/modals/sources_modal_inner.html
```diff
@@ -64,24 +66,30 @@ <h2 class="accordion-header">
...
     let lastGroup = null;
+    let lastDoc = null;
 
     let uniqueGroups = new Set();
     for (let item of accordionItems) {
       uniqueGroups.add(item.getAttribute("data-group"));
     }
-    if (uniqueGroups.size === accordionItems.length) return;
 
     accordionItems.forEach((item, index) => {
```
smcdonald-jus, line 76: Moderately thick solid lines to separate granular groups of sources, very thick dotted lines to separate documents

django/tests/chat/test_chat_views.py
```diff
@@ -852,7 +868,8 @@ def test_per_source_qa_response(client, all_apps_user):
     # Create a chat using the route to create it with appropriate options
     response = client.get(reverse("chat:qa"), follow=True)
     chat = Chat.objects.filter(user=user).order_by("-created_at").first()
-    chat.options.qa_answer_mode = "per-source"
+    chat.options.qa_granular_toggle = True
```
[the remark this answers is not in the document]
smcdonald-jus: no

django/chat/utils.py
```diff
@@ -490,13 +497,19 @@ def create_batches(iterable, n=1):
...
         for i, stream in enumerate(streams):
             try:
                 if stream["status"] == "running":
-                    final_streams[i] += next(stream["stream"])
-            except StopIteration:
+                    if isinstance(stream["stream"], Iterable):
+                        final_streams[i] += next(stream["stream"])
+                    else:
+                        partial_streams[i] = await stream["stream"].__anext__()
+                        final_streams[i] = partial_streams[i]
```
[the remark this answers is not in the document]
smcdonald-jus: This is...not true. Just double-checked and sure enough: `'async_generator' object has no atrribute '__iter__'`.

It's a moot point because this is actually a now-unused piece of logic from my attempt to do nested batching (which I've abandoned in this PR, but may be worth revisiting in the future), but still good to make note of this.

PR #445, opened 2025-07-24
add: text extractor size 

Rifahaziz:
Problem: Text extractor would reduce small images to the point that it became unreadable with human eye when pasted on the A4 white page during merging.

Solution:
-Trimmed invisible whitespaces on images.
-Added the options to have **small vs enlarged** images pasted on the A4 white page during merge option only.
-Enlarged option during merge allows the output to have bigger images.
-Small option during merge keeps the original small size.


smcdonald-jus:
Regarding the user options:

Would it be more intuitive if, instead of "Small/Enlarged", we had checkbox that toggled them, labelled with something like "Enlarge small images on page"?

Rifahaziz:
I agree, added as asked :)

PR #454, opened 2025-08-05
add: download all files as zip in text extractor

Rifahaziz:
**DO NOT MERGE  until Michel's PR is merged.**
Few things changed: 
-polling task was dependent on passing user request id, this is changed to checking for poll_tasks now as user request id is required to download all files as zip for each user. 
-zip download all icon available for non-merging case only as that produces multiple outputs
-test added


Rifahaziz:
@juleskuehn sure!, Ali found a bug here where the celery task is not being updated correctly when status is "FAILURE" for corrupted files. I'm working on fixing that atm. Thanks

django/text_extractor/templates/text_extractor/completed_documents.html
```diff
@@ -88,11 +88,10 @@
 {% elif swap %}
   <div id="warn-dont-leave-page" hx-swap-oob="true" class="d-none"></div>
 {% endif %}
-{% if show_download_all_button %}
-  <script>document.querySelector("#download-all-files").classList.remove("d-none");</script>
-{% endif %}
-{% if user_request_id %}
+
+{% if should_poll %}
```
Rifahaziz, line 92: polling condition based on should_poll and NOT user_request_id

PR #457, opened 2025-08-13
feat: custom query for keyword law search

smcdonald-jus:
Implements my fix to the LlamaIndex keyword search so we don't have to wait for the next round of requirement updates for that. Also (naively) applies keyword search to metadata so titles and section numbers are included. May still be dominated by vector similarity in unexpected ways, though.


smcdonald-jus:
Closing this for now to clear it out of the way. I will reopen it when we have more bandwidth to experiment with custom queries. It could also tie into an agentic approach that would reweight the metadata retriever in accordance with the query.

PR #459, opened 2025-08-13
add: pin and unpin chats

Rifahaziz:


Rifahaziz:
- [ ] for the pin icon (this is just aesthetic) use classes bi bi-pin-angle text-muted 
- [ ] don't show the pinned chat section if it is empty
- [ ] don't allow pinning an empty chat actually this is pretty hard to get right, and it's an edge case - who cares.
- [ ]  move pin button up one position (keep "delete" at the bottom)
- [ ] get rid of the heading "options" in the ... menu 
- [ ]  don't display in 2 places; just show in pinned if its pinned 
- [ ]  instead of refreshing the page (this may cause issues if a response is streaming, file is uploading etc while users perform the "pin" action) return HTMX swap of the #chat-history-list or #left-sidebar whichever is easier.

PR #463, opened 2025-08-14
feat: app tabs

juleskuehn:
image

### DIFF
--- a/django/locale/translation/translations.json
+++ b/django/locale/translation/translations.json
@@ -4726,5 +4726,9 @@
     "Maximum 2000 files / 300MB total": {
         "fr": "",
         "fr_auto": "Maximum de 2000 fichiers / 300 Mo au total"
+    },
+    "show app list": {
+        "fr": "",
+        "fr_auto": "Afficher la liste des applications"
     }
 }


django/otto/views.py
```diff
@@ -110,26 +111,23 @@ def get_categorized_features(user):
...
-            or request.user.homepage_tour_completed,
-        },
-    )
+    # Determine if the tour should be forced
+    force_tour = not request.user.homepage_tour_completed
+    tour_skippable = request.user.is_admin or request.user.homepage_tour_completed
 
+    # If an existing user is logging in and doesn't need the tour, redirect to the AI assistant
+    if request.session.pop("from_welcome", False) and not force_tour:
+        return redirect("chat:new_chat")
```
haitmo, line 120: I’m not entirely on board with this idea, mainly because the AI assistant is just one of the tools available on the platform. That said, I understand it’s the most widely used, so I’m not completely opposed to it.

haitmo (approved):

django/otto/views.py
juleskuehn: I understand that rationale. We could make it more "neutral" (but functionally the same really) by just returning to the last used app. Can adjust later based on feedback.

PR #470, opened 2025-08-20
feat: redirect to subsection when viewing law sources

smcdonald-jus:
In response to user feedback, have the "Go to source document" links in Legislation Search sources scroll down automatically to the relevant (sub)section, using the lims-id's which correspond to element id's in the HTML.

### DIFF
--- a/django/laws/views.py
+++ b/django/laws/views.py
@@ -71,7 +71,7 @@ def index(request):
 
 def source(request, source_id):
     source_id = urllib.parse.unquote_plus(source_id)
-    source_node = get_source_node(source_id)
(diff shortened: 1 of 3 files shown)


django/laws/views.py
```diff
@@ -92,10 +92,16 @@ def source(request, source_id):
             else None
         )
     law.url = get_law_url(law, lang)
+
+    if law.short_title == "THE CONSTITUTION ACTS, 1867 to 1982":
```
smcdonald-jus, line 96: Making a note of this, since I should verify that this logic _will_ always pick out the Constitution stuff.

PR #471, opened 2025-08-20
feat: documents count in library/folder

Rifahaziz:
Extra information added in library:

- total document count
- total failed count
- total stopped count
- total cost 
- total chunks


Rifahaziz:
Fixed: 
- Wrapped the fail/stop counts in a tooltip icon 
- fail/stop counts updating instantly with the correct number 
- "chunk(s)" word not disappearing anymore (i.e on updating from polling it is returning the word too instead of just the number)

haitmo (approved):

PR #474, opened 2025-08-25
fix: laws tables headings

amusbag:
Fix the laws tables headings by repeating the table header in each chunk to retain context.


juleskuehn (commented):
I'm not quite done reviewing because there are so many possible cases, but at a glance it looks good! I made some naming changes. Perhaps Shaun can test the behaviour too, making sure to look at several examples of Schedules.

PR #476, opened 2025-08-27
fix: access "data source status" as library viewer

smcdonald-jus:
https://dev.azure.com/BusinessAnalyticsCentre/BAC%20Agile/_workitems/edit/3320

Polling view now requires library viewing permission, not editing permission. Viewers will no longer get booted with an "unauthorized access" notification when viewing a document's status during processing.


amusbag (approved):
Reviewed and looks good.

PR #478, opened 2025-08-28
Chat dowload chat

mcusteau:


juleskuehn (dismissed):

juleskuehn (approved):
just changed this to modify print styles for now, so users can print to PDF with correct formatting. we can work on the proper PDF/DOCX export again later.

PR #479, opened 2025-08-28
add: sort and search in librarian

Rifahaziz:
librarian documents can now be:
- Sorted with date,name,type and chunks/size.
- Searched


haitmo (commented):
@Rifahaziz Polling resets the search filters, any chance we can fix that as well? Sorry I missed it earlier.

haitmo (approved):

PR #490, opened 2025-09-08
fix: show French-only names for Libraries and folders in English

smcdonald-jus:
https://dev.azure.com/BusinessAnalyticsCentre/BAC%20Agile/_boards/board/t/Otto/Projects?workitem=3333

### DIFF
--- a/django/otto/settings.py
+++ b/django/otto/settings.py
@@ -353,6 +353,8 @@ LOCALE_PATHS = [
     os.path.join(BASE_DIR, "locale"),
 ]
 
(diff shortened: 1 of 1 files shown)


django/otto/settings.py
```diff
@@ -353,6 +353,8 @@
     os.path.join(BASE_DIR, "locale"),
 ]
 
+MODELTRANSLATION_FALLBACK_LANGUAGES = ("en", "fr")
```
juleskuehn, line 356: Nice

PR #492, opened 2025-09-11
feat: chat search

Rifahaziz:
A search button added on the left side history bar that does the following: 

- searches chat titles 
- searches chat contents
- highlights a chat that has the content (if there are multiple matches in the content, it highlights one), should I change this?


Rifahaziz:
Keeping track of the improvements asked:

- The current chat should still be bolded.
- Pressing the "X" inside the search box should not only clear it, the bar should go away too. And the page URL should update to remove the query string about search.
-  When it highlights the chat message, (I really like this btw!) please highlight only the message-blob
- In the list of results, it would be better to have the details in a second line (small, lighter text) than in parentheses. I think it might be better just not to show it at all though.
-  When you clear the search results, your current chat isn't highlighted anymore.
- Why isn't there the "..." in the search results?
- After clicking a chat, when page loads there is a flash of the full list, then it shows the shorter results list.
- So if there are a lot of issues with the sidebar behaviour it might be better to have a modal for the search - the button would open the modal. The modal can have search results that are a bit longer (maybe have the preview of the relevant message there) like in ChatGPT.Then when you click the result, everything goes back to normal.
- 16. OR, you can try to figure how to make changes to the sidebar to resolve some of the issues. It seems like it is close and I like it overall.

juleskuehn (approved):
Looks great! I made some small tweaks to styling etc. and handling some edge cases (when you have an untitled chat open).

Rifahaziz:
I think it is good to go now,timeout solved the scrolling issue.  @juleskuehn

juleskuehn (approved):
Nice work! Made some minor tweaks. This is a great feature.

PR #495, opened 2025-09-16
fix: translation features & warnings

juleskuehn:
Requires running migrations and `reset_app_data presets`


amusbag (approved):
Looks good.

PR #497, opened 2025-09-19
fix: polling issues

haitmo:
1. The document status should now only be updated if it is being processed.
2. The race condition between user interaction with a document in the document list and polling should be fixed as well.
- I've testing it by both emulating a slow network connection and adding timeouts to the details page view function.

Could not reproduce the following issues:

> Sometimes the document pane itself does not update when the file is completed processing.
> Re-processing a PDF (from the document pane "Process document.." dropdown) sometimes shows an Error, which disappears when you click the document title in the folder pane.

They could be addressed in another pr when more details are provided


juleskuehn (approved):
I wasn't able to test it thoroughly, but it doesn't seem to break anything and the code makes sense. Thanks!

PR #502, opened 2025-09-22
fix: remove bug where nested lists could be rendered as code blocks

smcdonald-jus:
https://dev.azure.com/BusinessAnalyticsCentre/BAC%20Agile/_boards/board/t/Otto/Projects?workitem=3341

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ django/celerybeat-schedule
 django/venv/*
 webcrawler/venv/*
 django/staticfiles/*
+legal_harmonization*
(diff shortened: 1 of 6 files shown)


.gitignore
```diff
@@ -7,6 +7,7 @@ django/celerybeat-schedule
 django/venv/*
 webcrawler/venv/*
 django/staticfiles/*
+legal_harmonization*
```
smcdonald-jus, line 10: For my own convenience as I'm working on that

PR #503, opened 2025-09-23
fix: display time in user's timezone

haitmo:
Times should be displayed in the user's timezone. Data is still stored in UTC

image

_instead of 16:26_


juleskuehn (approved):
Works like a charm! Thanks.

PR #506, opened 2025-09-25
fix: disable async usage in OttoLLM to address bug 3344

jannable:
Application was experiencing ConnectionDoesNotExistError during vector database queries, particularly on first requests to new models. The error occurred when asyncpg connections were closed mid-operation during JSON codec setup, likely due to event loop handling issues in threaded environments combined with the recent switch from transaction to session pooling (bug #3344).

- Disabled use_async=True in both get_retriever() and _get_tree_summarizer() functions
- Resolves immediate connection stability issues
- Performance impact expected to be minimal for current workloads
- Allows continued operation while investigating proper async/pooling configuration
- Re-enable async mode once root cause is addressed

### DIFF
--- a/django/chat/llm.py
+++ b/django/chat/llm.py
@@ -318,7 +318,7 @@ class OttoLLM:
             similarity_top_k=top_k,
             num_queries=1,  # set this to 1 to disable query generation
             mode="relative_score",
-            use_async=True,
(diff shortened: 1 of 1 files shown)


jannable:
No longer needed since we resolved the original differently, by removing pgbouncer and using application connection pooling instead.

PR #507, opened 2025-09-27
feat: split embedding model demand between questions and documents

jannable:
This pull request updates our Azure OpenAI embedding deployment and application code to prevent heavy document processing from rate limiting the embedding service needed by users who are asking chat questions.

**What Was Changed**

Infrastructure:
- The Terraform now creates two deployments of the text-embedding-3-large OpenAI model:
  - One for user/questions (text-embedding-3-large-questions)
  - One for batch document ingestion (text-embedding-3-large-documents)
- The OpenAI quota (tokens-per-minute) is split between these two:
  - A percentage is reserved for chat questions (user requests)
  - The remainder is reserved for background document processing

Django Code:
- The OttoLLM embedding wrapper now accepts an embedding_deployment parameter.
- All code paths that handle user chat or Q&A are switched to use the "questions" deployment.
- All librarian/document-ingest code paths now use the "documents" deployment.
- The default embedding deployment in code is now the "questions" model, to favour real-time user responsiveness.

**Why This Was Done**
- When both chat and bulk document processing use the same embedding model deployment, a surge of ingestion jobs (like big uploads or library batch jobs) could exhaust the model’s rate limit. This would cause regular users' chat questions to be slowed down or even blocked.
- By splitting the quota at the Azure deployment layer and routing workloads to different deployments, we ensure that:
  - Day-to-day user chat and Q&A remains fast and responsive.
  - Background document processing cannot "starve" the interactive experiences, even under heavy ingestion load.

### DIFF
--- a/django/chat/llm.py
+++ b/django/chat/llm.py
@@ -95,12 +95,16 @@ class OttoLLM:
     "model" must match the name of the LLM deployment in Azure.
     """
 
+    # Since the questions deployment has less quota, use it by default for embeddings
+    # Document processing must be explicitly configured, which is appropriate since it's a specialized use case
+
     def __init__(
         self,
         deployment: str = settings.DEFAULT_CHAT_MODEL,
         temperature: float = 0.1,
         mock_embedding: bool = False,
         reasoning_effort: str = "minimal",
+        embedding_deployment: str = "text-embedding-3-large-questions",
     ):
         # Check if mock_llm is enabled via contextvar (for load testing)
         self.use_mock_llm = mock_llm_context.get(False)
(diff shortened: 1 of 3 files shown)


haitmo:
@jannable Do we want to track the "new" models in 2 new cost types?

jannable:
> @jannable Do we want to track the "new" models in 2 new cost types?

I don't think it's necessary since the cost is the same regardless of which instance of the same model is used. The split is only for quota management and to protect the interactive use cases from the batch processing. If you foresee another need to to have two cost types though, let's discuss.

juleskuehn:
> @jannable Do we want to track the "new" models in 2 new cost types?

This is already done basically - they are the same CostType, but each Cost object also has a "feature" property. For queries, the feature is "qa" and for loading it is "librarian" or "laws-loading" etc. As I recall.

django/chat/llm.py
```diff
@@ -95,12 +95,16 @@ class OttoLLM:
     "model" must match the name of the LLM deployment in Azure.
     """
 
+    # Since the questions deployment has less quota, use it by default for embeddings
```
juleskuehn: This comment is confusing; both comments be removed IMO.

juleskuehn (approved):
Nitpick about the comments, looks good though!

PR #517, opened 2025-10-10
fix: NLTK cache in Dockerfile, quieter logs, Entra user skip on blank email

jannable:
- Added NLTK download in Dockerfile for caching and avoid dynamic loading
- Added logic to only show healthz, user_cost, and notifications if there is a warning/error
- Added condition to skip Entra users if email is empty

### DIFF
--- a/django/otto/utils/logging.py
+++ b/django/otto/utils/logging.py
@@ -2,6 +2,7 @@ from typing import Any, Mapping, MutableMapping
 
 from django.dispatch import receiver
 
+import logging
 import structlog
 from django_structlog.signals import bind_extra_request_metadata
 from structlog.processors import CallsiteParameter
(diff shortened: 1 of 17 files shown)


django/otto/utils/logging.py
```diff
@@ -22,3 +23,17 @@ def merge_pathname_lineno_function_to_location(
     func_name = event_dict.pop(CallsiteParameter.FUNC_NAME.value, None)
     event_dict["location"] = f"{pathname}:{lineno}({func_name})"
     return event_dict
+
+
+class RaiseLevelForEndpointsFilter(logging.Filter):
```
haitmo, line 28: I'd like to stick to structtlog for those, but I'll adjust it in the other repo instead when we merge it all in

haitmo (approved):

